PR #6, opened 2015-04-23
Let recognizer connect to remote app.

clemenshelm:


DominikGuzei:
Build fails :wink:

clemenshelm:
Yeah, but it works locally. Had a look at it, seems it's because Codeship is using different versions of libraries. Not gonna fix that now, I suppose the problem will disappear at some point.

PR #12, opened 2016-08-09
Feature/vat id recognition

tamacodechi:
Sorry for the million commits 😅


lib/bill_recognizer.rb
```diff
@@ -47,8 +52,8 @@ def recognize
 
       Word.create(text: word_node.text, left: left, right: right, top: top, bottom: bottom)
     end
-    # puts Word.map(&:text)
-    # puts Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
+    #  puts Word.map(&:text)
+    puts Word.map { |word| "text: '#{word.text}', left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
```
clemenshelm: Could you comment it out again so we don't have massive amount of logs in production? We should replace this with a logger soon where we can set different log levels for different environments.

lib/detectors/vat_number_detector.rb
```diff
@@ -0,0 +1,41 @@
+require_relative '../word_list'
+require_relative '../models/vat_number_term'
+
+class VatNumberDetector
+
+  AUSTRIAN_VAT_REGEX = /ATU\s*\d{8}/
+  EU_VAT_REGEX = /EU\s*\d{9}/
+  LUXEMBURG_VAT_REGEX = /LU\s*\d{8}/
+  GERMAN_VAT_REGEX = /DE\s*\d{9}/
+  IRISH_VAT_REGEX = /IE\s*\d{7}[A-Z]/
```
clemenshelm: Why did you include the `\s*` in the regexes? IMHO they shouldn't be necessary since we're merging words without spaces.

spec/detectors/vat_number_detector_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    it "recognizes a VAT ID number where number is larger font", :focus do
+      create(:word, text: 'Wien', left: 2229, right: 2293, top: 301, bottom: 327)
+      create(:word, text: 'ATU', left: 2130, right: 2177, top: 339, bottom: 363)
+      create(:word, text: '37893801', left: 2196, right: 2323, top: 338, bottom: 365)
+      create(:word, text: 'EUR', left: 2505, right: 2553, top: 379, bottom: 404)
+
+      vat_numbers = VatNumberDetector.filter
+      expect(vat_numbers.map(&:to_s)).to eq ['ATU37893801']
+    end
```
clemenshelm: Please fix the intendation.

Gemfile
```diff
@@ -32,6 +32,9 @@ gem 'aws-sdk', '~> 2'
 gem 'sqlite3'
 gem 'sequel'
 
+# Valvat validates european VAT ID numbers 
+gem 'valvat', '~> 0.6.10'
+
```
clemenshelm: I think we should remove this since we're not using it yet.

clemenshelm:
Good job! 🐰

Gemfile
tamacodechi: Should we remove it completely or comment it out?

tamacodechi:
All done and the tests still pass! 🙌

PR #19, opened 2016-08-18
Feature/vat id validation

tamacodechi:
Had to go and add the EU VAT regex, or else it would pick up funny things like `EUR0` as a vat ID.

Thank you for all the help 🎉  Really happy this works!

### DIFF
--- a/lib/detectors/vat_number_detector.rb
+++ b/lib/detectors/vat_number_detector.rb
@@ -1,47 +1,43 @@
+require 'valvat'
 require_relative '../word_list'
 require_relative '../models/vat_number_term'
+require_relative '../logging.rb'
 
 class VatNumberDetector
+  include Logging
 
-  # AUSTRIAN_VAT_REGEX = /ATU\d{8}/
-  # EU_VAT_REGEX = /EU\d{9}/
-  # LUXEMBURG_VAT_REGEX = /LU\d{8}/
-  # GERMAN_VAT_REGEX = /DE\d{9}/
-  # IRISH_VAT_REGEX = /IE\d{7}[A-Z]/
-  VAT_NUMBER_REGEX = /[A-Z][A-Z][A-Z0-9]{6,12}/
+  VAT_REGEX = /[A-Z]{2}[A-Z0-9]{2,12}/
 
   def self.filter
-    find_vat_number(VAT_NUMBER_REGEX) # This one is run first because Amazon uses two VAT IDs and we need their LU ID.
-    # find_vat_number(GERMAN_VAT_REGEX) # This one runs second because when LU is not present we need DE
-    # find_vat_number(AUSTRIAN_VAT_REGEX)
-    # find_vat_number(EU_VAT_REGEX)
-    # find_vat_number(IRISH_VAT_REGEX)
+    detected_vats = find_vat_numbers(VAT_REGEX)
+    detected_vats.each do |term|
+      if Valvat.new(term.to_s).valid?
+        term.save
+      end
+    end
+
     VatNumberTerm.dataset
   end
 
-
   private
 
-  def self.find_vat_number(regex, after_each_word: nil)
+  def self.find_vat_numbers(regex, after_each_word: nil)
     term = VatNumberTerm.new(regex: regex, after_each_word: after_each_word)
     last_word = nil
 
     median_height = Word.map(&:height).sort[Word.count / 2]
-    Word.where('bottom - top <= ?', median_height * 1.15).each do |word|
+    terms = Word.where('bottom - top <= ?', median_height * 1.15).map do |word|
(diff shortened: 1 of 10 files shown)


lib/detectors/vat_number_detector.rb
```diff
@@ -1,41 +1,45 @@
...
 
-  def self.find_vat_number(regex, after_each_word: nil)
+  def self.find_vat_numbers(regex, after_each_word: nil)
     term = VatNumberTerm.new(regex: regex, after_each_word: after_each_word)
     last_word = nil
 
     median_height = Word.map(&:height).sort[Word.count / 2]
-    Word.where('bottom - top <= ?', median_height * 1.15).each do |word|
-      if term.exists? || (last_word && !word.follows(last_word))
+    term_complete = false
```
tamacodechi: Would it be useful for us to document the reason behind fixes like these somewhere? 📄 🔍

lib/detectors/vat_number_detector.rb
```diff
@@ -1,41 +1,45 @@
...
 
   def self.filter
-    find_vat_number(LUXEMBURG_VAT_REGEX) # This one is run first because Amazon uses two VAT IDs and we need their LU ID.
-    find_vat_number(GERMAN_VAT_REGEX) # This one runs second because when LU is not present we need DE
-    find_vat_number(AUSTRIAN_VAT_REGEX)
-    find_vat_number(EU_VAT_REGEX)
-    find_vat_number(IRISH_VAT_REGEX)
+    detected_vats = find_vat_numbers(VAT_REGEX)
+    detected_vats
+      .select { |term| term.text[EU_VAT_REGEX] || Valvat.new(term.to_s).valid? }
```
clemenshelm, line 15: `term.text[EU_VAT_REGEX]` looks very nice, I never thought of that. Still I'm wondering if the elegance doesn't lead to less clarity compared to `term.text =~ EU_VAT_REGEX`. What do you think?

lib/detectors/vat_number_detector.rb
clemenshelm: Hmm… I think we could use this method in all other detectors as well. I think the concept is much nicer now since now it does what it says: It finds vat numbers, nothing more. IMHO the saving step should be separate, like in this class.

I'd leave it for now but let's make a refactoring session next week where we clean this up, ok?

lib/detectors/vat_number_detector.rb
```diff
@@ -1,41 +1,45 @@
...
+  def self.find_vat_numbers(regex, after_each_word: nil)
     term = VatNumberTerm.new(regex: regex, after_each_word: after_each_word)
     last_word = nil
 
     median_height = Word.map(&:height).sort[Word.count / 2]
-    Word.where('bottom - top <= ?', median_height * 1.15).each do |word|
-      if term.exists? || (last_word && !word.follows(last_word))
+    term_complete = false
+    terms = Word.where('bottom - top <= ?', median_height * 1.15).map do |word|
+      if term_complete || (last_word && !word.follows(last_word))
```
clemenshelm: Oh, I just saw something. Could you do me a favor and try if it works to
- put `term.valid?` on this line instead of `term_complete`
- remove all the other `term_complete` stuff

I think that will work as well.

lib/models/price_term.rb
```diff
@@ -14,14 +14,9 @@ def initialize(attrs)
...
 
-  def before_create
-    @term_builder.pack!
-    self.text ||= @term_builder.text
-  end
-
   def add_word(word)
     @term_builder.add_word(word)
-
+    self.text = @term_builder.text
```
clemenshelm: Should be `@term_builder.extract_text` instead.

lib/models/date_term.rb
```diff
@@ -12,15 +12,11 @@ def initialize(attrs)
...
-  def before_create
-    @term_builder.pack!
-    self.text = @term_builder.text
-    self.first_word_id = @term_builder.words.first.id
-  end
-
   def add_word(word)
     @term_builder.add_word(word)
 
+    self.text = @term_builder.text
```
clemenshelm: Should be `@term_builder.extract_text` instead.

lib/detectors/vat_number_detector.rb
tamacodechi: Worked on the detection specs, running all the specs now 👍

lib/detectors/vat_number_detector.rb
tamacodechi, line 15: To me it's pretty clear, but I wrote it so I might be biased 😅  I think `=~ regex` is such a rubyism that it's commonly less clear, but Ruby is our friend so that might not be the case?

clemenshelm:
Good job! 🐰

PR #21, opened 2016-09-20
Feature/billing period detector

### LINKED ISSUE #71: No currency detected
Bills GGjBYXD5rNbJHdXC8.pdf has no currency result, although the currency is clearly stated on the bill.

tamacodechi:
Finally, a PR for this feature 👏 
Thank you for all the help! 😁 
Hope I didn't miss any indentations!

_Edit: This will probably take a bit long, I was about 30% done when I ran all the tests and there were 40 failing tests. I think it would be better to run the tests after each of the fixes to make sure everything stays intact._

**Fixes:**
- [x] **spec/acceptance/bill_recognition_spec.rb** - date period iso 
- [x] **lib/detectors/billing_period_detector.rb** - require_relative
- [x] **lib/detectors/billing_period_detector.rb** - setting text
- [x] **lib/boot.rb** - foreign keys
- [x] **lib/calculations/billing_period_calculation.rb** - require_relative
- [x] **lib/calculations/date_calculation.rb** - require_relative
- [x] **lib/bill_recognizer.rb** - invoice billing period
- [x] **lib/detectors/billing_period_detector.rb** - from > to
- [x] **lib/models/dimensionable.rb** - following.height
- [x] **spec/detectors/billing_period_detector_spec.rb** - unclear describe
- [x] **spec/models/dimensionable_spec.rb** - describe dimensionable

Closes #71


lib/bill_recognizer.rb
```diff
@@ -55,30 +59,35 @@ def recognize
...
-    vat_amount = prices.vat_amount
-
     date_words = DateDetector.filter
+    vat_number_words = VatNumberDetector.filter
+    billing_period_words = BillingPeriodDetector.filter
+    currency_words = CurrencyDetector.filter
+
+    billing_period = BillingPeriodCalculation.new(
+      billing_period_words
+    ).invoice_billing_period
```
clemenshelm: I don't see the value of naming it `invoice_billing_period`. We're only in the context of an invoice here, so it seems a little redundant to me. 😉

lib/boot.rb
```diff
@@ -40,6 +40,13 @@
   Integer :bottom
 end
 
+DB.create_table :billing_period_terms do
+  primary_key :id
+  String :text
+  Integer :from_id
+  Integer :to_id
```
clemenshelm: Could you use `foreign_key` here as described in http://sequel.jeremyevans.net/rdoc/files/doc/association_basics_rdoc.html? Advantage: Explicit reference to the primary key of another table.

lib/calculations/billing_period_calculation.rb
```diff
@@ -0,0 +1,14 @@
+require_relative '../detectors/date_detector'
+require_relative '../models/billing_period_term'
+require_relative '../models/word'
```
clemenshelm: Why do we need the `require_relative`s here? Seems like these classes don't get referenced in the file. If it's for the specs, please move the `require_relative`s to the spec file.

lib/calculations/date_calculation.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../detectors/date_detector'
+require_relative '../boot'
```
clemenshelm: Why do we need this `require_relative` here, but not on the `BillingPeriodCalculation`?

lib/detectors/billing_period_detector.rb
```diff
@@ -0,0 +1,26 @@
+require_relative '../../lib/boot'
+require_relative '../word_list'
+require_relative './date_detector'
```
clemenshelm: Are the `require_relative`s above really necessary?

lib/detectors/billing_period_detector.rb
```diff
@@ -0,0 +1,26 @@
...
+require_relative '../models/billing_period_term'
+require_relative '../models/date_term'
+
+class BillingPeriodDetector
+  def self.filter
+    Word.where(text: ["-", "bis"]).all.select do |term|
+      from = DateTerm.right_before(term)
+      to = DateTerm.right_after(term)
+
+      if from && to && from.text < to.text
```
clemenshelm: What does `from.text < to.text` do?

lib/detectors/billing_period_detector.rb
```diff
@@ -0,0 +1,26 @@
...
+        term =  BillingPeriodTerm.new(
+          text: from.text + ' - ' + to.text,
```
clemenshelm: Since `from` and `to` are both dependencies of `BillingPeriodTerm`, the assignment of the text should be in the `initialize` method (since you probably want the text to have this format for all `BillingPeriodTerm`s).

Also, the `text` attribute is supposed to represent the original text, so the dates should be separated by the original separator (`"-"` or `"bis"`).

lib/models/dimensionable.rb
```diff
@@ -6,4 +6,19 @@ def width
...
+  end
+
+  module ClassMethods
+
+    def right_before(current)
+      find {|previous| previous.right < current.left}
+    end
+
+    def right_after(current)
+      all.find {|following| (following.left > current.right) && (following.left - current.right) < (following.height * 1)}
```
clemenshelm: `following.height` instead of `(following.height * 1)` would be enough 😉

spec/acceptance/bill_recognition_spec.rb
```diff
@@ -75,6 +76,7 @@
     expect(bill_attributes[:amounts]).to eq [{total: 9487, vatRate: 20}]
     expect(bill_attributes[:invoiceDate]).to eq '2015-04-10'
     expect(bill_attributes[:vatNumber]).to eq 'ATU45011703'
+    expect(bill_attributes[:billingPeriod]).to eq '01.03.2015 - 31.03.2015'
```
clemenshelm: This is inconsistent. `bill_attributes[:invoiceDate]` is formatted as `YYYY-MM-DD` whereas the `bill_attributes[:billingPeriod]` uses `DD.MM.YYYY`. Please also use `YYYY-MM-DD` here.

There is an ISO specification for representing date ranges (https://de.wikipedia.org/wiki/ISO_8601#Zeitspannen), so I recommend that we use it. For this example the format would be `2015-03-01/2015-03-31`.

spec/detectors/billing_period_detector_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    create(:word, text: '10.03.2016', left: 1026, right: 1288, top: 1148, bottom: 1190)
+    create(:word, text: 'Gesamtbetragfbruttoj', left: 1348, right: 1948, top: 1240, bottom: 1301)
+    create(:word, text: '1000,-', left: 2145, right: 2291, top: 1244, bottom: 1293)
+
+    DateDetector.filter
+    billing_periods = BillingPeriodDetector.filter
+    expect(billing_periods.map(&:text)).to eq ['07.02.2016 - 10.03.2016']
+  end
+
+  it "does not recognize an invoice date that is the same as part of the billing period as part of the billing period" do
```
clemenshelm: I don't understand this description.

spec/models/dimensionable_spec.rb
```diff
@@ -0,0 +1,134 @@
+require_relative '../../lib/boot'
+require_relative '../../lib/models/dimensionable'
+require_relative '../../lib/models/date_term'
+require_relative '../../lib/models/word'
+require_relative '../support/factory_girl'
+require_relative '../factories'
+
+describe DateTerm do
```
clemenshelm: Actually it should describe `Dimensionable`.

clemenshelm (changes requested):
Good job! Just a few little questions and fixes 😉

lib/bill_recognizer.rb
tamacodechi: Good point! I think I initially did it to keep conformity with invoice date 😅

spec/detectors/billing_period_detector_spec.rb
tamacodechi: I can't think of a better way to phrase it... Basically if there is a date that is repeated, eg.:
1.11.2011,
1.11.2011,
-,
20.11.2011.

It will ignore the first one, and still save that as a standalone date. There was an error where this repeating date would be saved as part of a billing period.

clemenshelm: Ok. But you aren't testing here that the standalone date isn't part of the billing period, are you? Maybe a description could be something like `"it doesn't consider other dates as part of the billing period"`?

tamacodechi: Sounds good, it was more specific to dates that are the same as dates from the billing period, but I think your suggestion for the description is clearer :)

lib/detectors/billing_period_detector.rb
```diff
@@ -0,0 +1,23 @@
+require_relative '../boot'
+require_relative '../word_list'
+require_relative '../models/billing_period_term'
+require_relative '../models/date_term'
+
+class BillingPeriodDetector
+  def self.filter
+    Word.where(text: ['-', 'bis']).all.select do |term|
```
clemenshelm: The purpose of `select` is to create a resulting array with filtered values (see http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-select).

If you only want to iterate over an array, use `each` instead.

clemenshelm (changes requested):
Excellent! Just one tiny remark :)

Issue #23, opened 2016-09-22
Lint Ruby


clemenshelm:

tamacodechi:
### Update:

I'm done with an [older version](https://github.com/clemenshelm/chillbill-recognizer/commit/c262cf7b18e3aa5d9f17681f31a5a514f1aeb7ac):
`47 files inspected, no offenses detected` 

I had started this before the PR for #21 was done reviewing, so I'll leave this as is until #21 and #22  gets merged, and then pull in the changes - then leave the _alllllll the code base pretty_ ✨  
I'd like to add rubocop to the CI (but I think that's part of #26 ) so that it'll yell at us for sure if there are syntax errors (I mean we should all be running it anyway before doing PRs but we're all human here, right? ಠ_ಠ  ... Right?) and at least throw warnings for other bad things we do.

tamacodechi:
Can/Should I add a couple of lines on rubocop to the read me in case anyone who joins in dev of the recognizer who isn't familiar with it (esp. the other cops) can just find info there?

_Sorry for all the comments, testing is slow._ ʱªʱªʱª (ᕑᗢूᓫ∗)

clemenshelm:
Would be nice :)

PR #28, opened 2016-10-03
created iban-detector

### LINKED ISSUE #22: IBAN detection
Extract IBANs from invoices. Make sure that IBANs aren't detected as UID numbers and vice versa.

IBANs contain two check digits, so we should be able to determine if it is a valid IBAN or not (https://en.wikipedia.org/wiki/International_Bank_Account_Number#Check_digits).

There are already Ruby gems for this validity check.

Test bills:
- [x] xfkjpC4oR694RLaQ4
- [x] 249d36tzwFBWjsrev
- [x] 22uKq5q9pa39DRfNB
- [x] 28YRpHtS7R3qwEfMR
- [x] 2SmZNKp2sb5j9H5Tr
- [x] 28fdHSL6w4YkD5DK6
- [x] 27XQiYsRgfuY8Qgkr
- [x] 36reqGNW2tEufnXJx
- [x] 6jwujKzunQ9Ni7EBg
- [x] 4T67zD7Fc9gygc7Qy
- [x] 3ajvp4ZMyx3FhFF2z
- [x] 7YoE7K6hcRvMv3o4r
- [x] 2MXM3jkEPyDLd7yf2

Tr4in:
Added iban detector, closes #22


tamacodechi:
Will this get merged before Black Bird, and so need to be covered by #23 ? confusedparrot

clemenshelm:
Yes, would be nice.

lib/bill_recognizer.rb
```diff
@@ -54,7 +58,8 @@ def recognize
       Word.create(text: word_node.text, left: left, right: right, top: top, bottom: bottom)
     end
     # logger.debug Word.map(&:text)
-    logger.debug Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
+    #logger.debug Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
+    puts Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
```
tamacodechi: - [x]  Could you comment this out again? We don't want it to have it in the running version ;)

lib/detectors/iban_detector.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class IbanDetector
+  AT_IBAN_REGEX = /AT[0-9]{18}/
+  DE_IBAN_REGEX = /DE[0-9]{20}/
+  HR_IBAN_REGEX = /HR[0-9]{19}/
+
+  def self.filter
+    iban_detected = find_iban(AT_IBAN_REGEX)
+    iban_detected = find_iban(DE_IBAN_REGEX)
+    iban_detected = find_iban(HR_IBAN_REGEX)
```
tamacodechi: Could we use one regex that accepts two characters at the beginning and a range of numbers from 18-20, or would this produce too many false positives? It would be good to just have to call find_iban once, right? :)

spec/acceptance/bill_recognition_spec.rb
```diff
@@ -325,4 +339,118 @@
...
+    expect(bill_attributes[:ibanNumber]).to eq "HR0923600001101675865"
+  end
+
+  it 'recognizes the bill 2MXM3jkEPyDLd7yf2' do
+    retriever = SpecCacheRetriever.new(file_basename: '2MXM3jkEPyDLd7yf2.pdf')
+    recognizer = BillRecognizer.new(retriever: retriever)
+
+    bill_attributes = recognizer.recognize
+    expect(bill_attributes[:ibanNumber]).to eq "AT761967500100912501"
+  end
```
tamacodechi, line 16: Did all of these tests fail initially? 

We don't usually add acceptance tests unless they fail because they're not providing any new scenarios to test.

spec/models/iban_term_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_relative '../../lib/boot'
+require_relative '../../lib/models/iban_term'
+
+describe IbanTerm do
+  it 'recognizes Austrain IBAN correctly'  do
```
tamacodechi: Tiny typo: Austrian

lib/detectors/iban_detector.rb
```diff
@@ -0,0 +1,36 @@
+require 'ostruct'
+require 'bigdecimal'
```
tamacodechi: What are we using these for here?
(Genuine question! I haven't needed them myself yet)

tamacodechi (changes requested):
Looks really good and clean - just a couple of things to tweak! 👍  Can't wait to see this merged!

spec/acceptance/bill_recognition_spec.rb
Tr4in: This test does not fail - this means I have to remove it?

tamacodechi: Well, the whole point is that we create an acceptance test that provides data: 
If the test fails, we write unit test and write/fix our methods. 
If it doesn't fail, the features that make it work have already been tested/written because of another test that is similar - so we don't need it and can remove it 😉

lib/detectors/iban_detector.rb
Tr4in: If I use one universal regex for these three then the recognizer does not recognize DE ibans anymore.

lib/detectors/iban_detector.rb
Tr4in: Deleted! :)

lib/bill_recognizer.rb
```diff
@@ -58,8 +58,8 @@ def recognize
       Word.create(text: word_node.text, left: left, right: right, top: top, bottom: bottom)
     end
     # logger.debug Word.map(&:text)
-    #logger.debug Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
-    puts Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
+    logger.debug Word.map { |word| "text: #{word.text}, left: #{word.left}, right: #{word.right}, top: #{word.top}, bottom: #{word.bottom}" }
```
tamacodechi: Could you comment this out too? We don't need this one in the running version either 😉

tamacodechi (changes requested):
Just one teeny tiny thing and then it's perfect 🎉  Great job @Tr4in :)

tamacodechi:
@Tr4in could you take a look at this again? The conflicting files are because master has changed in some files that you changed as well. 

Have you dealt with merge conflicts before? You're going to have to merge master onto your branch and make it so that your work is in the new version, while keeping the changes that were done to master.

Here's a nice article on merge conflicts: [boop](https://www.git-tower.com/learn/git/ebook/en/command-line/advanced-topics/merge-conflicts) If you need any help lemme know 😉

lib/bill_recognizer.rb
```diff
@@ -93,7 +100,8 @@ def recognize
     {
       amounts: [total: total, vatRate: vatRate],
       invoiceDate: invoice_date,
-      vatNumber: vat_number
+      vatNumber: vat_number,
+      ibanNumber: iban_number
```
clemenshelm: IBAN already stands for _International Bank Account Number_, so `ibanNumber` is kind of redundant 😄 

Please use just `iban`.

clemenshelm (changes requested):

lib/bill_recognizer.rb
Tr4in: Changed

tamacodechi:
Please run `rubocop` and see if there are any offenses.  They will probably mostly be lines that are too long in the test data so feel free to look at the others and style it similarly eg.

```
Term.create(
    text: 'sdfsdf',
    left: 21231,
    right: 123123,
    top: 231,
    bottom: 1231231
)
```

You can also run `rubocop -a` to autofix any problems like needing single quotes instead of double quotes.

clemenshelm (approved):

spec/calculations/iban_calculation_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require_relative '../../lib/boot'
+require_relative '../../lib/calculations/iban_calculation'
+require_relative '../../lib/models/iban_term'
+
+describe IbanCalculation do
+  before(:each) do
+    IbanTerm.dataset.delete
```
tamacodechi: This too?

lib/models/iban_term.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+require 'sequel'
+require_relative './term_builder'
+require_relative './dimensionable'
+
+# TODO: unit test
```
tamacodechi: You already wrote unit tests, so I don't think this comment's necessary :)

spec/calculations/iban_calculation_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require_relative '../../lib/boot'
+require_relative '../../lib/calculations/iban_calculation'
+require_relative '../../lib/models/iban_term'
```
tamacodechi: Should this be in spec helper along with the others? What do you think? 😁

spec/detectors/iban_detector_spec.rb
```diff
@@ -0,0 +1,186 @@
+# frozen_string_literal: true
+require_relative '../../lib/boot'
+require_relative '../../lib/detectors/iban_detector'
+require_relative '../../lib/models/word'
```
tamacodechi: boot and word are in the spec helper, so you don't need to require them in any spec file 😉  They're loaded up before any tests run

spec/detectors/iban_detector_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+require_relative '../support/factory_girl'
+require_relative '../factories'
+
+describe IbanDetector do
+  before(:each) do
+    Word.dataset.delete
+    IbanTerm.dataset.delete
+  end
```
tamacodechi: The word dataset is already cleared by spec helper, so you can remove it here :)

spec/models/iban_term_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+require_relative '../../lib/boot'
```
tamacodechi: Already required by the spec helper so you don't need to require it here 😉

lib/bill_recognizer.rb
```diff
@@ -110,6 +128,9 @@ def recognize
       customer_vat_number: @customer_vat_number
     ).vat_number
 
+    iban = IbanCalculation.new(iban_words).iban
+    # image_file.close
```
tamacodechi: Do we need this here before and after the currency calculation runs?

lib/detectors/iban_detector.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require_relative '../models/iban_term'
+
+class IbanDetector
+  AT_IBAN_REGEX = /AT[0-9]{18}/
+  DE_IBAN_REGEX = /DE[0-9]{20}/
+  HR_IBAN_REGEX = /HR[0-9]{19}/
+
+  class << self
```
tamacodechi: This was done to actually make the filter function private, right? It works just as well, but I think we should keep the code base looking kind of uniform so maybe it would be better to declare private within filter? That or change all the other detectors to be like this one.

tamacodechi (changes requested):
Looks really great! 🎉 

I tried running it on my machine to see if requiring the detector on the calculator was actually necessary and interestingly enough I get an error, can't run the tests on this branch.

It tells me that the running version's bundler is older so I have to update, but when I try to via docker-compose I get `Could not find rake-11.3.0 in any of the sources` . I've only gotten this on this branch, and now that I pulled in the newest changes, because when we worked together on it last week it was fine. I imagine that the version you had pushed did not have the updated Gemfile.lock

I would say go back to the older version on the branch, since we have this version on all branches. That or update the running version 😉  Then you'll be able to run the tests too :D

tamacodechi:
Awesome, it's great! 😁 
Could you merge in master to your branch? It should then be totally done I think! 👍

PR #35, opened 2016-10-13
lint all the things, add rubocop, remove unecessary acceptance tests, and set spec helper to detect CI ENV

### LINKED ISSUE #23: Lint Ruby

### LINKED ISSUE #26: Fix CI for recognizer
At the moment we don't see if a recognizer version works correctly (= all the tests pass) because our CI setup doesn't run the tests correctly. This makes approving features a hassle.
- [ ] Only run unit tests, only one acceptance test

`Remove all failing tests and leave one perfectly passing test`
- [x] Fail if there are focussed specs

``` ruby
  if ENV['CI']
    config.before(:example, :focus) { raise "Should not commit focused specs" }
  else
    config.filter_run :focus
    config.run_all_when_everything_filtered = true
  end 
```
- [x] Include Rubocop.

`rubocop`

tamacodechi:
## Things to note:
-  **Many**  things have changed and only some have changed to follow Rubocop's nitpicks:

> eg. only one regex is wrapped by %r{ } because Rubocop wanted it that way - if the others were styled the same the cop would get angry
- **There are still 20 offences.** These are found in `.rubocop_todo.yml` They are all things that have to be refactored, so I have broken them down by method and description and will put them up as issues with the refactoring label
- **Running** `rubocop -a` **breaks a test!!** It removes a comma from a word list in one of the price detector specs: `'finds prices that consist of 2 words'` which causes the test to fail.
- Rubocop insists that **line 53 of bill_recognizer.rb** is too long (it is!) and makes us split it up. Because of the only way it which it can be split (ie. in the middle of the string) you get this message when you run the acceptance test: `sh: 2: -l: not found` The test still passes and runs, but yeah...

Closes #23 and fixes #26


clemenshelm (approved):
Sooo nice! I also learned some new things 😄

PR #55, opened 2016-10-25
24 due date recognizer

### LINKED ISSUE #24: Recognize due date
Only if it's recognized as date already.

tamacodechi:
closes #24


lib/calculations/due_date_calculation.rb
```diff
@@ -0,0 +1,26 @@
...
+    @words = words
+  end
+
+  def due_date
+    return nil if @words.empty?
+
+    # Find due date label
+    due_word = Word.where(text: %w(Due Zahlungstermin))
+
+    due_date_label = Word.right_after(due_word.first) if due_word
```
clemenshelm: In what case is `due_word` falsy? It's only a Sequel query, so it should always be defined.

lib/models/dimensionable.rb
```diff
@@ -18,9 +18,19 @@ def right_before(current)
...
+          (following.left - current.right) < (following.height * 10) &&
+          following.bottom == current.bottom
+      end
+
+      # Finds word right after a little above or below
+      all.find do |following|
+        (following.left > current.right) &&
+          (following.left - current.right) < (following.height * 10.145) &&
+          ((following.bottom <= (current.bottom * 1.02)) &&
+          ((current.bottom * 0.95) <= following.bottom))
```
clemenshelm: Multiplying `current.bottom` doesn't make a lot of sense: If the word is closer to the bottom of the page, the product will be significatly larger (because the bottom value is a lot larger). I'd use `height` instead as this stays constant on the whole page.

clemenshelm: Why have you defined `all.find…` twice? The first one won't have any effect now as the result is just discarded.

clemenshelm: Can you please write specs for this method?

clemenshelm (changes requested):
Oh yeah! 😄

lib/calculations/due_date_calculation.rb
tamacodechi: Whoops! My bad! I intended it for if due_word was empty so it wouldn't run .right_after on nil. Added a test for this (because the one that returns nil is when there are no words contained, not "no due dates") and fixed 👍

lib/models/dimensionable.rb
tamacodechi: Please let me know if this makes sense 😅  

The reason I'm going through `all.find` twice is because I need to iterate over the words twice:
Once to find if there are any words on the same line, and if there are none, then another time to find if there are any words slightly above or below. 

I can't do this as a nested loop or condition because then _it will just take whichever term came first_. I don't want this because we explicitly need to see if there are any terms on the same line first, and then _if there are none_ find ones slightly below or above.

Is there another way I can solve this? This is why I had originally built a `on_same_line` method.

clemenshelm (approved):

PR #63, opened 2016-11-04
53 create deployment command

### LINKED ISSUE #52: Add a version to the recognizer
The recognizer should have a version number for each new version which gets deployed. This version number should automatically be generated in the deployment process and be sent with each recognition result so we can keep track in the web app which results were produced with which recognizer version.

### LINKED ISSUE #53: Create deployment command
At the moment it takes several commands to deploy the recognizer. There should be a single command that takes care of the whole deployment.

This command could be running a shell script, a rake task, a capistrano task or whatever fits.

tamacodechi:
closes #53 and closes #52 

Yay 🎉  Now let's hope nobody tinkers with the version.yml file hahahah

Update: WAAAAIT lemme lint it!
Update: Ok, all done 😅


Rakefile
```diff
@@ -73,3 +74,40 @@ def process(bill_kind, &bill_proc)
...
+      docker build -t recognizer-repo .
+
+      docker tag recognizer-repo:latest 175255700812.dkr.ecr.eu-central-1.amazonaws.com/recognizer-repo:latest
+
+      docker push 175255700812.dkr.ecr.eu-central-1.amazonaws.com/recognizer-repo:latest"
+end
+
+desc 'Restart task on ECS'
+task :restart_task => [:push_image] do
+  system_task = "tasks=$(aws ecs list-tasks --cluster ChillBill)
```
clemenshelm: - [x] Should probably also contain the region, otherwise an environment variable needs to be set. `--region eu-central-1`.

Rakefile
```diff
@@ -73,3 +74,40 @@ def process(bill_kind, &bill_proc)
...
+                runningtask=$(echo $tasks | tr -d \"\ {}\" | cut -d \"[\" -f2 | cut -d \"]\" -f1)
+                runningtask=$(sed -e 's/^\"//' -e 's/\"$//' <<<\"$runningtask\")
+                echo $runningtask"
+  running_task = `#{system_task}`
```
clemenshelm: - [x] I have no idea what this does. Why not just use Ruby?

``` ruby
tasks = `aws ecs list-tasks --cluster ChillBill`
running_task = tasks.match(/\/([^\/]+)$/)[1]
```

> Fix this regex so it doesn't pick up the end of the string. (Maybe length of the ID?)

Rakefile
```diff
@@ -73,3 +74,40 @@ def process(bill_kind, &bill_proc)
...
+
+  sh "aws ecs stop-task --cluster arn:aws:ecs:eu-central-1:175255700812:cluster/ChillBill --task #{running_task}
+
+      aws ecs start-task --cluster arn:aws:ecs:eu-central-1:175255700812:cluster/ChillBill --task-definition arn:aws:ecs:eu-central-1:175255700812:task-definition/ecscompose-recognizer:32 --container-instances arn:aws:ecs:eu-central-1:175255700812:container-instance/f8742655-f231-48ed-ab0d-a2aa92d94117"
```
clemenshelm: - [x] This contains the current revision number (32). When we publish a new revision of the task, the deployment script will still use the old version, so effectively we won't be able to run new versions of the task. Is there an easy way to sove this? (Maybe passing no revision number or `"latest"` works?

> List all and find the one with the biggest number

clemenshelm: - [x] Is it required that you pass the container instance in here? Looks quite redundant as the instance is associated with the cluster anyway.

> Use non ARN command

Also here in the `run-task` command there is no reference to the instance: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_AWSCLI.html#AWSCLI_run_task

Rakefile
```diff
@@ -73,3 +74,40 @@ def process(bill_kind, &bill_proc)
...
+
+      docker build -t recognizer-repo .
+
+      docker tag recognizer-repo:latest 175255700812.dkr.ecr.eu-central-1.amazonaws.com/recognizer-repo:latest
+
+      docker push 175255700812.dkr.ecr.eu-central-1.amazonaws.com/recognizer-repo:latest"
+end
+
+desc 'Restart task on ECS'
+task :restart_task => [:push_image] do
```
clemenshelm: - [x] Why does restarting require pushing the image?

clemenshelm (changes requested):
Very nice, but still room for improvement 😄

clemenshelm (approved):
Yeah 👍

clemenshelm:
Please resolve the conflicts, then you can merge it 😄

PR #80, opened 2016-11-11
Make vat ID detector return an empty data set if there are no words

### LINKED ISSUE #75: NoMethodError in VatNumberDetector
When there are no words found at all, the VatNumberDetector crashes:

```
NoMethodError: undefined method `*' for nil:NilClass
/myapp/lib/detectors/vat_number_detector.rb:27:in `find_vat_numbers'
```

Bill: PYdefyzCHkSp9atMY.jpg (empty Image)

tamacodechi:
closes #75

### DIFF
--- a/lib/bill_recognizer.rb
+++ b/lib/bill_recognizer.rb
@@ -104,7 +104,6 @@ class BillRecognizer
     #        bottom: #{word.bottom}
     #        "
     #      }
-    
     price_words = PriceDetector.filter
     logger.debug price_words.map { |word|
       "PriceTerm.create(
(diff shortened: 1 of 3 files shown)


lib/bill_recognizer.rb
```diff
@@ -104,7 +104,7 @@ def recognize
     #        bottom: #{word.bottom}
     #        "
     #      }
-
+
```
clemenshelm: No whitespace, please 😄

clemenshelm: Doesn't Rubocop lint this?

clemenshelm (changes requested):

lib/bill_recognizer.rb
tamacodechi: Strange thing is that it allowed it and it wouldn't pass the CI tests if it didn't. I also spotted this when I was creating the PR but I thought it was harmless since Rubocop didn't mind. I can remove it 😉

clemenshelm (approved):

PR #82, opened 2016-11-16
81 due date label detection

### LINKED ISSUE #81: Write due date label detection

tamacodechi:
Closes #81

The internet is terrible so I ended up having to use my phone's internet after a long battle :\ 

@clemenshelm Is it okay as "DueDate:" ? I know we prefer to have representations of the actual text. I'm thinking the reason it won't allow "Due Date:" is that the term builder doesn't add a space after last_word, but messing with the term builder sounds like an awful idea.

The reason the due date labels are not stored in a variable or anything is because we're not using it YET. We'll need it in the due date calculation refactor :)


lib/detectors/due_date_label_detector.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+require_relative '../models/due_date_label_term'
+
+class DueDateLabelDetector
+  DUE_DATE_LABELS =
+    /(?:^|(?<= ))(Zahlungstermin|DueDate:|Zahlungsziel:)(?:(?= )|$)/
```
clemenshelm: What are these groups at the beginning and at the end? `(?:^|(?<= ))` and `(?:(?= )|$)`?

clemenshelm: Please use "Due Date:". You can use `after_each_word: end_word_with_space` like in the `DateDetector`. You even added this option to your `find_due_date_labels` function 😉

spec/detectors/due_date_label_detector_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      :word,
+      text: '14 May 2015',
+      left: 534,
+      right: 650,
+      top: 1480,
+      bottom: 1519
+    )
+
+    due_date_labels = DueDateLabelDetector.filter
+    expect(due_date_labels.map(&:to_s)).to eq ['DueDate:']
```
clemenshelm: 'Due Date:'

clemenshelm (changes requested):
As you assumed, I'd like to have a "Due Date:" 😉

clemenshelm (approved):

PR #94, opened 2016-11-22
76 unknown file extension returns true error

### LINKED ISSUE #76: NoMethodError in BillRecognizer
When a bill file is not a supported type, the `BillImageRetriever` returns `true` instead of a reference to the file (on line 38). The bill recognizer expects the returned object to have a `path` method, though. This causes the error

```
NoMethodError: undefined method `path' for true:TrueClass
/myapp/lib/bill_recognizer.rb:55:in `recognize'
```

Bill: xcaEpkmTauDsZz9fk.p7s

tamacodechi:
Hope this is the solution we were looking for :v:

Aaalso, just noticed how long ago I created this branch. this really is not an indication of how long this took to do. Maybe I'm just worrying too much but I actually started on 3 issues at once because I thought they would be super quick fixes (oh boy was I wrong!) so it's super misleading how long ago I started working on this... Just in case you were using this as a metric to see how long a fix like this takes 😉 

Closes #76


lib/bill_image_retriever.rb
```diff
@@ -5,6 +5,14 @@
 require 'aws-sdk'
 require_relative './logging.rb'
 
+class UnprocessableFileError < StandardError
+  attr_reader :extension
+  def initialize(extension, message = 'Unprocessable file type: ')
+    @extension = extension
+    super(message + extension)
+  end
```
clemenshelm, line 13: Nice!

clemenshelm (approved):
Don't worry, be happy! 😄

PR #95, opened 2016-11-23
Refactor/due date calculation

tamacodechi:
That was surprisingly a lot less work than I anticipated 😅 
(But also this is due to the fact that I split up the tasks instead of doing one huge PR)


lib/calculations/date_calculation.rb
```diff
@@ -18,18 +18,7 @@ def invoice_date
...
-    return nil unless due_word.any?
-    due_date_label = Word.right_after(due_word.first)
-    date_after_label = if due_date_label.nil? || due_date_label.text != 'Date:'
-                         DateTerm.right_after(due_word.first)
-                       else
-                         DateTerm.right_after(due_date_label)
-                       end
-    date_after_label.to_datetime
+    return nil if DueDateLabelTerm.empty?
+    DateTerm.right_after(DueDateLabelTerm.first).to_datetime
```
clemenshelm, line 22: 100.000 times nicer 😄

spec/calculations/date_calculation_spec.rb
```diff
@@ -145,17 +146,17 @@
       bottom: 1519
     )
 
+    DueDateLabelDetector.filter
     due_date_calculation = DateCalculation.new(
-      DateTerm.dataset
+      DueDateLabelTerm.dataset
```
clemenshelm: Do we even need the argument here? In fact we access both `DateTerm` and `DueDateLabelTerm` directly in the calculation without using this argument.

clemenshelm (changes requested):
Cool! But what I'm missing here: How is the due date label term found? It should have an own detector, shouldn't it?

spec/calculations/date_calculation_spec.rb
tamacodechi: We get:
``` 
ArgumentError:
       wrong number of arguments (given 0, expected 1) 
```

tamacodechi:
It does have it's own detector :) The due date calculation just uses the terms found by the due date detector.

clemenshelm (approved):

PR #96, opened 2016-11-23
91 ft recognized as eur

### LINKED ISSUE #91: Ft. recognized as EUR
In bill bsg8XJqLBJSt2dXeH.pdf the currency is states as "Ft." which is the abbreviation of hungarian forints. The currency is recognized as EUR, though. Probably, because "EUR" is mentioned at the top of the page.

tamacodechi:
It works as a solution for now 🎉 
Proximity to total amount would be nice eventually 🤔 

Closes #91


clemenshelm (approved):
Good job 🐥

PR #99, opened 2016-11-24
Explicit region and considering no running tasks.

### LINKED ISSUE #97: Make deployment more robust

clemenshelm:
Closes #97


tamacodechi (approved):
Looks great 👍  I hadn't even considered that when there's no running tasks we wouldn't be able to start us a task. Thank you for fixing this! 🎉

PR #102, opened 2016-11-28
93 reprocess old bills

### LINKED ISSUE #93: Reprocess old bills
Subscribe to the new publication of bills to reprocess, but give these Sidekiq jobs lower priority, so new bills are always processed first.

tamacodechi:
Closes #93 
Is this alright? I'm kind of feeling around the dark since I can't really test it. I'd love some input so I can fix it further.


Rakefile
```diff
@@ -27,42 +27,19 @@ task process: :setup_processing do
...
-    end
+    id = recognition_result.delete :id
+    logger.info ["result for bill #{id}:", recognition_result].map(&:to_s)
+      .map(&:yellow).join(' ')
+    meteor.write_detection_result(id, recognition_result)
   end
 end
 
-def process(bill_kind, &bill_proc)
+def process(bill_kind, &bill_proc, queue)
```
clemenshelm: Order of attributes is incorrect, `queue` is the second argument.

lib/hub.rb
```diff
@@ -30,12 +31,13 @@ def run
       )
 
       meteor.subscribe("admin.bills.#{@bills}")
+      meteor.subscribe('admin.bills.toReprocess')
```
clemenshelm: Why?

lib/hub.rb
```diff
@@ -30,12 +31,13 @@ def run
...
 
       meteor.collection('bills')
             .on(:added) do |id, bill|
         logger.info "bill #{id} was added: #{bill}"
         bills[id] = bill
-        RecognitionWorker.perform_async id, bill[:imageUrl]
+        RecognitionWorker.perform_async id, bill[:imageUrl], :queue
```
clemenshelm: Why are you passing the `:queue` symbol?

Rakefile
```diff
@@ -17,7 +17,7 @@ end
 
 desc 'Process unprocessed bills'
 task process: :setup_processing do
-  process(:unprocessed) do |recognition_result, _bill, meteor|
+  process(:unprocessed, :unprocessed) do |recognition_result, _bill, meteor|
```
clemenshelm: It's not obvious what the parameters mean. Maybe use named parameters so it's clear what's happening here.

config/sidekiq.yml
```diff
@@ -1,2 +1,6 @@
 ---
 :concurrency: 1
+
+:queues:
+  - unprocessed
+  - reprocess
```
clemenshelm: Inconsistent naming (particip vs. verb)

clemenshelm (changes requested):
In general that's how it should be, comments inline 😄

lib/hub.rb
tamacodechi: Do I not need to subscribe to the publication of bills to be reprocessed that the web app sends, or are they all under @bills? 🤔

clemenshelm (approved):

lib/hub.rb
clemenshelm: `@bills` is the name of the subscription, so this line is unnecessary.

sidekiq.rb
```diff
@@ -20,7 +20,8 @@
 class RecognitionWorker
   include Sidekiq::Worker
 
-  def perform(id, bill_image_url)
+  def perform(id, bill_image_url, queue)
+    sidekiq_options queue: queue
```
clemenshelm: Did you check that this works inside of the method? I only saw examples where this was defined on class scope.

clemenshelm (changes requested):

sidekiq.rb
tamacodechi: I also thought this but I can't see another way to get the queue value to be able to set the option

clemenshelm: Ok, but did you check that it works? I don't want the recognition to stop working because of this line 😉

clemenshelm: Look what I found :)
http://www.rubydoc.info/gems/sidekiq/Sidekiq/Client#enqueue_to-class_method

tamacodechi: Awesome! So I can use something like 
```        
Sidekiq::Client.enqueue_to(
  :queue, RecognitionWorker, id, bill[:imageUrl]
)
```
. Since I need to do this before we call the worker and for each bill that is pulled from the subscription, would I do this in the rake task? (in which case I can pass the queue directly by name instead of in a variable)

Rakefile
```diff
@@ -71,15 +48,15 @@ def process(bill_kind, &bill_proc)
...
+  hub = Hub.new(bills: bill_kind, config: config, queue: queue)
   hub.run(&bill_proc)
 end
 
 desc 'Increment recognizer version number'
 task :increment_version do
   require 'YAML'
   data = YAML.load_file "lib/version.yml"
-  data["Version"] += 1
+  data["versionNumber"] += 1
```
clemenshelm: I think this shouldn't be here, right?

clemenshelm (changes requested):
Just 1 tiny detail :)

clemenshelm (approved):

PR #105, opened 2016-12-01
71 no currency detected

### LINKED ISSUE #71: No currency detected

tamacodechi:
Closes #71 
Didn't really create test data because it was something in how the recognizer works and that would mean creating more acceptance tests. It works on the bill that brought up the problem and our original acceptance test :v:


lib/bill_recognizer.rb
```diff
@@ -154,30 +154,33 @@ def recognize
...
-    vat_rate =
-      if sub_total.nonzero?
-        (vat_total * 100 / sub_total).round
-      else
-        0
-      end
+    amounts = [
+      total: 'Not detected',
+      vatRate: 'Not detected'
+    ]
```
clemenshelm: That's not good. Why not make it an empty array and insert an element in the `unless` block?

clemenshelm (changes requested):

lib/bill_recognizer.rb
```diff
@@ -157,30 +157,30 @@ def recognize
...
-        0
-      end
+    amounts = []
+
+    unless net_amount.nil?
+      # Adapt recognition result to application schema
+      # TODO: Let price calculation produce required format
+      sub_total = net_amount * 100
+      vat_total = vat_amount * 100
+
```
tamacodechi, line 171: This blank line is a problem isn't it? 😆

clemenshelm (approved):

Issue #70, opened 2016-11-10
Incorrect currency detected


clemenshelm:
In invoices mMHiT2b3C5fgYqBzY.pdf and WHPRwa2CSgNDDrsx9.pdf USD is recognized, but the bill's currency is EUR.

tamacodechi:
WHPRwa2CSgNDDrsx9.pdf is fixed but mMHiT2b3C5fgYqBzY.pdf has an OCR issue where a rogue $ sign is recognized that is not on the actual bill.

PR #106, opened 2016-12-01
70 incorrect currency detected

tamacodechi:
This does not close the issue #70 because it's an OCR issue. A random $ gets detected that's not on the bill, but once this gets fixed, this PR should fix the other problem, which is that Eur is not recognized yet when it's written as Euro (fixed in this PR).


clemenshelm:
Rubocop failed

PR #108, opened 2016-12-05
107 stop running task fix

### LINKED ISSUE #107: Stop running task in deployment broken

clemenshelm:
Closes #107 

Already deployed to production.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -99,7 +99,7 @@ task :restart_task do
   tasks = `aws ecs list-tasks --cluster ChillBill --region eu-central-1`
   running_tasks = tasks.match(/task\/(\w+\W\w+\W\w+\W\w+\W\w+)/)
   if running_tasks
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -99,7 +99,7 @@ task :restart_task do
   tasks = `aws ecs list-tasks --cluster ChillBill --region eu-central-1`
   running_tasks = tasks.match(/task\/(\w+\W\w+\W\w+\W\w+\W\w+)/)
   if running_tasks
-    sh "aws ecs stop-task --cluster ChillBill --task #{running_tasks[0]} --region eu-central-1"
+    sh "aws ecs stop-task --cluster ChillBill --task #{running_tasks[1]} --region eu-central-1"
```
tamacodechi, line 102: That's interesting, how come we want the second item?

clemenshelm: The first item is the match of the whole Regexp (including "task/"). The second item is the first matched group. Yes, confusing :)

PR #115, opened 2016-12-08
Speed up currency detection.

### LINKED ISSUE #114: Investigate bills timing out
On the recognizer a the recognitizon of the majority of bills run extremely wrong and times out eventually. Investigate, what's the problem.

clemenshelm:
Closes #114 

This sped up the currency detection for a bill from 69 seconds (!) to 5 seconds.

### DIFF
--- a/lib/detectors/currency_detector.rb
+++ b/lib/detectors/currency_detector.rb
@@ -2,7 +2,7 @@
 require_relative '../models/currency_term'
 
 class CurrencyDetector
-  EUR_SYMBOLS = %w(EUR € EURO).freeze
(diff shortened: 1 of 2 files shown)


tamacodechi:
What a cool fix! 🌟  This is so useful for the detectors that have loads and loads of possible terms (soon the due date label detector). I'm surprised Rubocop was okay with the word lists, I try and stay away from them because Rubocop tends to hate them.

clemenshelm:
I think Rubocop just complains if you surround them with something other than parentheses. Other than that I haven't had issues so far.

PR #118, opened 2016-12-12
Limit number of words per term.

### LINKED ISSUE #117: Still a lot of bills timing out
Investigate why there are still a lot of bills taking very long to be recognized. Fix the issue.

clemenshelm:
Closes #117

### DIFF
--- a/lib/detectors/currency_detector.rb
+++ b/lib/detectors/currency_detector.rb
@@ -22,13 +22,13 @@ class CurrencyDetector
     CurrencyTerm.dataset
   end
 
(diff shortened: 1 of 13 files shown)


tamacodechi (approved):
Amazing! 👏
I have a bug in the relative date detector where it would return the detected word as many times as there are words on the file, so I was actually looking at and wondering about ```word.each do``` just just now! This is a really good fix, and if ever a word limit needs raising it looks very easy to change the max word count!

clemenshelm:
Thanks :)

PR #120, opened 2016-12-13
101 relative date detection

### LINKED ISSUE #77: NoMethodError in due date detection
When there is no due date found after a due date label, there is a `NoMethodError`:

```
NoMethodError: undefined method `to_datetime' for nil:NilClass
/myapp/lib/calculations/date_calculation.rb:33:in `due_date'
```

Bill: ZqMX24iDMxxst5cnP.pdf

### LINKED ISSUE #101: Relative date detection
We will need this for invoices that contain words such as "prompt" (which will then allow me to finish up #77)

I think it would be better to discuss a little before I start building something completely unnecessary. Should this be a set of words that get detected in it's own detector and then get used in the date calculation file to calculate relations to other dates on the bill?

Would love your input @clemenshelm

tamacodechi:
Yay 🎉  I'm pretty happy with this one. It looks like it will be relatively simple to build upon! 

I'm a little unsure about a couple of things, that I hope it's okay to get some confirmation on:

**lib/bill_recognizer.rb**
- [x]  Line 164 : Should this have the if invoice_date confirmation as well...? Or rather, should I somehow makes this all a little more robust, because if there is no invoice date it will pass nothing to the due date calculation, as I understand it?
- [x] Would it be a good idea to/can I add ``` if DueDateLabelTerm.any?``` to the end of ```due_datetime = DateCalculation.new(date_words).due_date(invoice_date)```

**lib/calculations/date_calculation.rb**
- [x] Should there be a case for the conditional where there is neither a date nor a relative date after the due date label and therefore it just returns nil? 

Closes #101 and closes #77


lib/bill_recognizer.rb
```diff
@@ -155,9 +158,11 @@ def recognize
     iban = IbanCalculation.new(iban_words).iban
 
     currency = CurrencyCalculation.new(currency_words).iso
-    due_datetime = DateCalculation.new(date_words).due_date
+    due_datetime = DateCalculation.new(date_words).due_date(invoice_date)
```
clemenshelm: In fact, you don't need to pass the `invoice_date` here, you can just call the `invoice_date` method inside the `DateCalculation`. You also don't need to make a new `DateCalculation` instance here since it's already instantiated at line 146.

clemenshelm:
Regarding point 2: At the moment not since there's no "real-life" test backing it up.

clemenshelm:
Same goes for point 3.

clemenshelm (changes requested):

lib/bill_recognizer.rb
tamacodechi: Hmm. That's true about making the new ```DateCalculation```, but about ```invoice_date``` wouldn't that mean I'm calculating the invoice date from the ```due_method``` and then not returning it?

lib/detectors/relative_date_detector.rb
```diff
@@ -0,0 +1,33 @@
...
+class RelativeDateDetector
+  RELATIVE_WORDS = /(prompt)/
+
+  def self.filter
+    find_relative_words(RELATIVE_WORDS)
+    RelativeDateTerm.dataset
+  end
+
+  def self.find_relative_words(regex, after_each_word: nil)
+    private
```
clemenshelm, line 13: `private` within a method doesn't make sense. I know that this is part of almost all Detectors, but that doesn’t make it more correct :)

clemenshelm (changes requested):
Very nice! Just one tiny detail 😄

lib/detectors/relative_date_detector.rb
tamacodechi: Hmm, I think this may have been a Rubocop autocorrect. Should I change it on all the detectors?

clemenshelm: Yes, please. Just remove it everywhere.

clemenshelm: Oh, and please resolve the conflicts with master. Thank you 😄

tamacodechi: Sounds good, but I should still make all those methods private, right? Not just remove t hat keyword ;)

clemenshelm: Would be nice, but the `private` keyword doesn't work for class methods. Here are some suggestions on that: http://stackoverflow.com/questions/4952980/creating-private-class-method

clemenshelm (approved):

PR #122, opened 2016-12-13
Use RMagick to read PDFs.

### LINKED ISSUE #119: Process PDFs with RMagick directly

clemenshelm:
Closes #119

### DIFF
--- a/lib/image_processor.rb
+++ b/lib/image_processor.rb
@@ -5,7 +5,15 @@ class ImageProcessor
   include Magick
 
   def initialize(image_path)
-    @image = Image.read(image_path)[0]
+    begin
+      original_image = Image.read(image_path)[0]
+      min_dimension = [original_image.page.width, original_image.page.height].min
+    ensure
+      original_image&.destroy!
+    end
+    # This will give us an image with at least 3000px on each dimension
+    density = 220_000.0 / min_dimension
+    @image = Image.read(image_path) { self.density = density }[0]
   end
 
   def apply_background(color)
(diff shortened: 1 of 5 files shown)


lib/image_processor.rb
```diff
@@ -44,13 +53,15 @@ def improve_level
   end
 
   def trim
-    @image.fuzz = '99%'
+    @image.fuzz = '80%'
```
tamacodechi, line 56: Just out of curiosity, how come we're changing this?

lib/image_processor.rb
```diff
@@ -5,7 +5,16 @@ class ImageProcessor
...
   def initialize(image_path)
-    @image = Image.read(image_path)[0]
+    begin
+      original_image = Image.read(image_path)[0]
+      page = original_image.page
+      min_dimension = [page.width, page.height].min
+    ensure
+      original_image&.destroy!
+    end
+    # This will give us an image with at least 3000px on each dimension
```
tamacodechi, line 15: Awesome, will this potentially fix those weird errors we get where the bottom of a bill is cut off?

tamacodechi (approved):
Looks great and cuts away so much of the retriever code! 👍

lib/image_processor.rb
clemenshelm: Probably not.

lib/image_processor.rb
clemenshelm: But maybe this will.

clemenshelm: Because the acceptance test didn't work anymore. All of a sudden it cut off a lot more. The `fuzz` is kind of the tolerance for trimming, so now it needs a lower tolerance.

PR #126, opened 2016-12-14
123 fix due date calculation

### LINKED ISSUE #123: Due date calculation crashes when no due date found after label
Bills:

- [x] a8sPrtNYneSzxram9.pdf
- [x] xELTimruvWAv5HvKY.pdf
- [x] qZZbW4R2w5EMcq83D.pdf – relative: „sofort fällig”
- [x] yiaGswKDskiLNkafN.pdf
- [x] RmHboa947xBm7bjby.pdf – relative: „sofort fällig”
- [x] JdySXxNkr7ycJTJMT.pdf – relative: „sofort fällig”
- [x] xAkCJuSGM8A4ZGoSy.pdf
- [x] hKL25ASsBcaZ47Hy4.pdf
- [x] eKiitgrZHzPfyycNL.pdf
- [x] ZyfTxiR2seqtm3Q8o.pdf
- [x] LfZKWFL9eGvrrv58F.pdf
- [x] xkGhBQwr44R4cWNoJ.pdf
- [x] YkDPFtWHFfMMSeQuo.pdf
- [x] 8ZEQJF6zqu2bfQsxW.pdf
- [x] x8BAy7QvjyMWGYLZC.pdf
- [x] 3EagyvJYF2RJhNTQC.pdf
- [x] 3tePSrWob4PiANTgK.pdf
- [x] 45SrgXjZqL5eyR8go.pdf – Date recognized incorrectly
- [x] 465YqPwgHMCTARESh.pdf
- [x] 2Wc6pX3mxJvq4tzLs.pdf
- [x] 2k9YJcQyas4z76gZf.pdf
- [x] 8gabRmC6wFbjguMFu.pdf
- [x] 98xawHkEKTwMSqdLa.pdf
- [x] A7QtBbTFZ6c3PbW4H.pdf
- [x] AD2SgGHBE3gF3pMkG.pdf
- [x] AZj9XptPKJmGMjB8N.pdf
- [x] Av8bBp6YkJabcAMcd.pdf
- [x] Ax9gh29o8wk3XWozf.pdf
- [x] BgohLR8rb75Wi5QYe.pdf

clemenshelm:
Closes #123 

Fixed some other things that I found along the way 😄


spec/detectors/billing_period_detector_spec.rb
```diff
@@ -203,4 +203,50 @@
       '2015-11-30'
     )
   end
+
+  it 'does prefers from dates closer to the separator' do
```
tamacodechi: Tiny nitpick: it does prefers is a bit weird, would it be okay to just put prefers?

lib/models/dimensionable.rb
```diff
@@ -14,16 +14,29 @@ def self.included(base)
...
+      end
     end
 
     def right_after(current)
       all.find do |following|
         (following.left > current.right) &&
-          (following.left - current.right) < (following.height * 15) &&
-          ((following.bottom <= (current.bottom + following.height)) &&
-          (following.bottom >= (current.bottom - following.height)))
+          (following.left - current.right) < (following.height * 19) &&
```
tamacodechi, line 25: Since this multiplier keeps changing should we note somewhere what case it is that made us change it to that value specifically?

lib/calculations/date_calculation.rb
```diff
@@ -15,15 +15,17 @@ def invoice_date
...
     return nil if DueDateLabelTerm.empty?
-    due_date = DateTerm.right_after(DueDateLabelTerm.first)
-    if due_date
-      DateTerm.right_after(DueDateLabelTerm.first).to_datetime
-    else
+    due_date_term = DateTerm.right_after(DueDateLabelTerm.first)
+    due_date_term ||= DateTerm.below(DueDateLabelTerm.first)
+    due_date = due_date_term&.to_datetime
+
+    due_date || begin
```
tamacodechi, line 28: I adore all of this fix. Just for me to learn, is there a reason why we use ||= instead of doing it on the same line with ||?

lib/detectors/date_detector.rb
```diff
@@ -65,8 +74,13 @@ def find_dates(regex, after_each_word: nil, max_words: nil)
...
           last_word = word
 
-          term.save if term.valid?
+          if term.valid?
+            term.save
+            affected_words += term.words
+          end
         end
+
+        affected_words
```
tamacodechi, line 83: I just want to make 100% sure I understand this.  

At the end of each call of ```find_dates``` we are returning an array that contains a copy of all the instance variables of term, which will be the term in its completed form (ie. not cut up into separate words).._. right?_

spec/models/dimensionable_spec.rb
```diff
@@ -29,50 +29,76 @@
...
-      top: 773,
-      bottom: 809
-    )
+      DateTerm.create(
+        text: '31.03.2015',
+        left: 832,
+        right: 1038,
+        top: 773,
+        bottom: 809
+      )
```
tamacodechi, line 64: Ridiculous question, but why is github highlighting little blocks where indentation already existed? Just if you know.

tamacodechi (approved):
Love the fixes, sorry for all the questions 😅

spec/detectors/billing_period_detector_spec.rb
clemenshelm: I guess I rewrote it somehow and must have left "does" 😁

lib/models/dimensionable.rb
clemenshelm: I think this "note" is the spec :)

lib/calculations/date_calculation.rb
clemenshelm: No, it works either way. But it would be too long on one line and this way I don't have to wrap one line of code over multiple lines. But it's really just preference.

lib/detectors/date_detector.rb
clemenshelm: What do you mean by "all the instance variables of term"? An instance variable is something like `@words`.

What's happening here is: We collect all the _words_ that make up the _terms_. So if there's a date term `23. September 2017` and another term `24.8.2016` there would be four words in the `affected_words` afterwards.

The reason is that on line 24 I remove the words that got recognized for the first regex from the word list that gets processed with the following regexes to avoid that the same words get recognized (wrongly) with another regex.

spec/models/dimensionable_spec.rb
clemenshelm: It's indented one level deeper now because I wrapped it with `describe '#right_before'` :)

PR #127, opened 2016-12-14
Read only first page of PDF.

### LINKED ISSUE #125: Out-of-memory error
Bill KigeeggEb3TYRWgih.pdf causes an out-of-memory error. In the past these errors were caused by RMagick when preprocessing bills. So this might be a good starting point to look at.

clemenshelm:
Closes #125 

One hour research, one line of code 😀


tamacodechi (approved):
Awesome fix though, well worth that research time I bet 👍

PR #133, opened 2016-12-20
86 due date recognized as invoice date

### LINKED ISSUE #86: Due date recognized as invoice date
In bill cAfvoH3zHjxmp88Ls.pdf the due date (2016-11-11) is listed as invoice date, probably because it's the first date. The correct invoice date is 2016-10-28.

tamacodechi:
Closes #86 
Hope this is the kind of solution we were looking for :) It's getting a bit bulky and I keep thinking if there isn't a way to stop the repetition found in the term files.


lib/detectors/invoice_date_label_detector.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+require_relative '../models/invoice_date_label_term'
+
+class InvoiceDateLabelDetector
+  INVOICE_DATE_LABELS =
+    /(Rechnungsdatum:)/
```
clemenshelm: The parentheses shouldn't be necessary.

spec/models/invoice_date_label_term_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+require_relative '../../lib/boot'
+
+describe InvoiceDateLabelTerm do
+  it 'detects a due date label correctly' do
```
clemenshelm: Probably `invoie date label` 😉

clemenshelm (changes requested):
Very nice! Only 2 details.

clemenshelm (approved):

PR #136, opened 2016-12-21
89 super future dates

### LINKED ISSUE #89: Super future invoice date
In bill p6eYWvCeXyeApYB7t.pdf the invoice date gets recognized as 4510-10-24. I'm not quite sure why, probably the text recognition for the billing period is wrong. If that's the case, we can't do much about it (yet). Still it would be a good idea to prevent invoice dates so far ahead in the future from being recognized.

tamacodechi:
closes #89 
I did bad thing and wrote the solution without testing because the only thing there would be to test would be to find this mysterious super future date and write an ``` it 'ignores super future dates' ``` test, but actually this super future date was a result of the detection stringing together an address and VAT ID. I did actually test it with the acceptance test and it does pick up the correct date now but I've reverted the acceptance test to what it should be. Is this a problem?


clemenshelm:
So the original test data didn't fail anymore?

tamacodechi:
The interesting thing is when I tried to recreate the problemsticresult on a clean branch off master I would get the correct result, meaning no problem. Even though this solution is healthy, should we delete this PR since the original data does not fail? Or can someone try to recreate this problem?

clemenshelm:
Interesting. I queried the DB for bills with recognized invoice date > year 2100 for the current recognizer version and I found these 36:

 - [x] Sqc9ixBz4g8mDCdJK.pdf
 - [ ] Q6Ty75NPSstBn9H5X.jpg
 - [ ] xiMDtAhcRMfhuwe7p.pdf
 - [ ] Ja388fziFPYE6Zx2j.pdf
 - [ ] nZvhLsJT9WpbyCy79.pdf
 - [ ] hNTSQ35hzF52BcQ5r.pdf
 - [ ] TYyh3RuRkjdQrkoG4.pdf
 - [x] wkGmT7JL8gYCSKDpc.pdf
 - [x] e8vEddezb52ASZDso.pdf
 - [ ] w4zBvgMZ4vt8KrjcX.pdf

 - [x] 3KdmxTXLCTMdyeduw.pdf
 - [x] 4XABPzur6atKRvCJc.pdf
 - [ ] 2gnRr9ThK6a2PiWxf.pdf
 - [ ] 5xRzts8XFDi9QApWC.pdf
 - [x] 6GiWj7tehdimjcpeA.pdf
 - [x] 6rd46tSX8YpnTvmke.PDF
 - [x] 7ZRtQo8xe82sDfxjX.PDF
 - [ ] 9neE7YxvRqMxSAcca.pdf
 - [ ] AeQsG5gs6ruumF9jp.pdf
 - [x] Az8gyBmNHxduLekrC.pdf

 - [x] Fe8C86yBXME5F5Mt3.pdf
 - [ ] GHPeccvjtKo8eK3eM.pdf
 - [x] KkhkRpms7eWhyBYyX.pdf
 - [x] sMMSHJyCdCKvCZ7ra.jpg
 - [x] ok2HWMRekbNNjXFox.jpg
 - [ ] YYKh5PH2QgAP3TBGc.pdf
 - [ ] xH55APEvc5cbji5kW.pdf
 - [ ] 8k2R9g7oM7tgpiLyH.pdf
 - [ ] bjxDxzZz4PC3F5D4w.pdf
 - [ ] BqhGLvxk4zaiHBW5w.pdf

 - [ ] FsZPCR9omH4SAvJ7m.pdf
 - [ ] vcQYP9hASFNztnwEq.pdf
 - [ ] vAdmaP46nWvic3DuY.pdf
 - [ ] CXCZWHDaNA2e4LLFp.pdf
 - [ ] HW9Shprq4xqdZ3jkP.PDF

Could you try to reproduce it on one of them? 😄

clemenshelm:
I love having a lot of data at hand 👑

tamacodechi:
Thank you! It's the best to have so much data at hand 🙆

tamacodechi:
I'm actually gonna test a couple and see if they need different changes to the regex because already master broke one of the results so I'll see if the others need updating too.

tamacodechi:
Okay! So the verdict:

- The ticked ones are all fixed now!
- The rest have OCR issues (So now they return nil for invoice date 🙌 )
- CXCZWHDaNA2e4LLFp.pdf and 5xRzts8XFDi9QApWC.pdf are not read properly because they're in the wrong orientation
- w4zBvgMZ4vt8KrjcX.pdf is a terrible scan and barely picks up anything
- AeQsG5gs6ruumF9jp.pdf has the date written in by hand, and it's a wonder a human can read it so I don't blame the poor recognizer
- sMMSHJyCdCKvCZ7ra.jpg and ok2HWMRekbNNjXFox.jpg are duplicates of each other
- 6rd46tSX8YpnTvmke.PDF and 7ZRtQo8xe82sDfxjX.PDF are duplicates of each other
- this was a nightmare 🙃

lib/detectors/date_detector.rb
```diff
@@ -5,14 +5,15 @@
 class DateDetector
   months = ((1..9).map { |n| "0?#{n}" } + (10..12).to_a).join('|')
   days = ((1..9).map { |n| "0?#{n}" } + (10..31).to_a).join('|')
-  SHORT_PERIOD_DATE_REGEX = /(?:^|[^+\d])((?:#{days})\.(?:#{months})\.\d+)/
+  SHORT_PERIOD_DATE_REGEX =
+    /(?:^|[^+\d])((?:#{days})\.(?:#{months})\.([2][0]\d{2}|[1]\d))/
```
clemenshelm: You don't need to put the numbers into square brackets. Square brackets are only necessary when you have multiple values to choose from, e.g. `[a-z2]` would allow lowercase letters from a to z and the digit 2.

So in this case the year regex could look like `(20\d{2}|1\d))`.

Then the other question is why you use either a number with 4 digits and arbitrary last two digits or a number with 2 digits which must start with `1`.

I'd suggest the following: `((?:20)?\d{2})`. This makes only the `20` optional. The `?:` at the beginning of the "century group" just means "Don't capture this group in the results".

lib/detectors/date_detector.rb
```diff
@@ -5,14 +5,15 @@
...
   SHORT_SLASH_DATE_REGEX = %r{((?:#{days})/(?:#{months})/\d{2}$)}
+  LONG_HYPHEN_DATE_REGEX = /((?:#{days})-(?:#{months})-[2][0]\d{2}$)/
```
clemenshelm: Also no square brackets necessary.

lib/detectors/date_detector.rb
```diff
@@ -5,14 +5,15 @@
...
   SHORT_ENGLISH_DATE_REGEX = /((?:#{days})-(?:Oct)-\d{4}$)/
   LONG_SLASH_DATE_REGEX = %r{((?:#{days})/(?:#{months})/\d{4}$)}
   FULL_GERMAN_DATE_REGEX = /(\d+\. (?:März|April|September|Oktober|Dezember) \d+)/
   FULL_ENGLISH_DATE_REGEX = /(\d+ (?:March|May|October) \d+)/
-  LONG_HUNGARIAN_DATE_REGEX = /\d{4}\.(?:#{months})\.(?:#{days})/
-  LONG_HYPHEN_DATE_REGEX = /((?:#{days})-(?:#{months})-\d{4}$)/
+  LONG_HUNGARIAN_DATE_REGEX = /[2][0]\d{2}\.(?:#{months})\.(?:#{days})/
```
clemenshelm: Also no square brackets necessary.

clemenshelm (changes requested):
Nice! You could tweak your regexes a little, though. 😉

lib/detectors/date_detector.rb
tamacodechi: Ahhh! Okay I couldn't find ```?:``` anywhere and it's so prevalent in our code, so thank you 💃 

Well I only really did what the data needed. The bills that showed super future dates that were fixed by limiting the 20xx, never actually needed fixing the decade, whereas the future dates that showing up with only a two digits were fixed by limiting the decade.

Considering the above, would something like ``` ((?:20)?1\d) ``` be best?

clemenshelm: Ok. The problem with the decade is that the decade will be over soon 😄 

But we can leave it for now if the test data requires it. Another solution would be for the short date to require that the next letter is NOT a digit.

tamacodechi: This is something I thought of as well, but I don't think it will be a problem until the iffy transition between 2019 and 2020 when people might still be sending in old bills, right? (but relying on the fact that we will update this each decade is messy, I agree).

As in the filter does not pick up more numbers right? Otherwise if it dismisses anything with a number following could be messy if there's two dates next to each other, like when they list due date, invoice date etc... in columns.

clemenshelm:
In order to make the tests work on Codeship, my pull request must be merged first so the whole new Codeship-with-Docker thing is set up.

clemenshelm (approved):

PR #138, opened 2016-12-23
Fix invalid date exception in various cases.

### LINKED ISSUE #128: Invalid date found
```
ArgumentError: invalid date
/myapp/lib/models/date_term.rb:54:in `strptime'
```

Bills:

- [x] gvjmW6sfEHvaXiWco.pdf
- [x] Y4SjLTSKps4hpdyZe.pdf
- [x] rq6SXCKCeXWB7Dhdh.png
- [x] yfKrfpuxv35Siiszk.pdf

clemenshelm:
Closes #128 

@Althaire please put detections into the detector, not into the calculator 😉 Thanks!


lib/detectors/date_detector.rb
```diff
@@ -10,7 +10,7 @@ class DateDetector
   SHORT_ENGLISH_DATE_REGEX = /((?:#{days})-(?:Oct)-\d{4}$)/
   LONG_SLASH_DATE_REGEX = %r{((?:#{days})/(?:#{months})/\d{4}$)}
   FULL_GERMAN_DATE_REGEX = /(\d+\. (?:März|April|September|Dezember) \d+)/
-  FULL_ENGLISH_DATE_REGEX = /(\d+ (?:March|May|October) \d+)/
+  FULL_ENGLISH_DATE_REGEX = /(\d{2} (?:March|May|October) \d{4})/
```
tamacodechi, line 13: I didn't touch the regexes with +s because I assumed that this meant they could have 1 or 2 digits for the day and 2 or 4 digits for the year. Would it be wiser to change it to accept 1 or 2 for the day and 2 or 4 for the year? ...Since it was built with + originally, presumably for a reason  😆

lib/detectors/billing_period_detector.rb
```diff
@@ -10,13 +10,28 @@ def self.filter
...
       next unless from && to
-      term =  BillingPeriodTerm.new(
+      term = BillingPeriodTerm.new(
         from: from, to: to
       )
 
       term.save
     end
 
+    if BillingStartLabelTerm.any? && BillingEndLabelTerm.any?
```
tamacodechi, line 20: Hmmm... I had a bit of trouble wrapping my head around it, but we should "put detections in the detectors" because we are assigning words to "to" and "from" in the original detector but in the extension (with the BP labels) I had assignments to "to" and "from" in the calculation?

I understand that they are essentially the same because both are searching through recognized words for billing periods' "to" and "from" but I feel like the original detector runs though all the recognized words in the document and filters out the BPs, whereas this extension with the labels runs through already detected words and calculates out the BP "to" and "from". So it detecting the BP or calculating it? And should the calculator only apply the end logic (ie. the method that puts the variable's content in the format we want)?

Just want to clear it up so I build correctly in the future and wrap my head around the core concepts of the recognizer correctly 😅

lib/bill_recognizer.rb
```diff
@@ -214,6 +214,7 @@ def preprocess(image_path)
          .deskew
          .normalize
          .trim
+         .improve_level
```
tamacodechi, line 217: I couldn't find this in the docs 😅  What does this do? Is it like an auto correction of the level channels?

lib/detectors/billing_period_detector.rb
```diff
@@ -10,13 +10,28 @@ def self.filter
...
+
+      billing_period_end = DateTerm.right_after(
+        BillingEndLabelTerm.first
+      )
+
+      BillingPeriodTerm.create(
+        from: billing_period_start,
+        to: billing_period_end
+      )
+    end
```
tamacodechi, line 33: Should we have this section before the the one that filters from hyphen and bis? This is a much more accurate recognition since it relies on explicit labels that only apply to the BP, and the hyphen/bis filter could pick up something erroneous first and apply that as the first BP, even when there are labels on the document.

tamacodechi (commented):
Looks good all around 🙋  Just a couple of questions!

tamacodechi (changes requested):
Please run ```rubocop --auto-gen-config```
Rubocop doesn't like the recognizer length is different than the one described on the todo file 😅

lib/detectors/date_detector.rb
clemenshelm: I would leave it for now. Everything else would be an assumption which is not backed with test data yet. But yes, probably that's how it's gonna turn out in the end. I just like to see it fail beforehand so we've got data for a new test.

lib/detectors/billing_period_detector.rb
clemenshelm: It's a special case here because the detector is actually a higher-level detector which builds on data already detected by another detector.

For me what differentiates a detector from a calculation is:

- The detector creates terms from already existing data. Each term has a detector, so the `BillingPeriodDetector` creates `BillingPeriodTerm`s and so on.
- The calculation chooses from the underlying terms the appropriate one. So the `BillingPeriodCalculation` would choose the right `BillingPeriodTerm` to be used as the billing period of the invoice.

A strong indicator that what you were doing was actually a detection was that the spec for the calculation said `it "detects …`.

lib/detectors/billing_period_detector.rb
clemenshelm: I think it doesn't matter, both sections can insert `BillingPeriodTerm`s. So no matter in what order you execute them, by the end of the method you'll return a dataset which contains results from both sections. Figuring out the right `BillingPeriodTerm` is subject to the calculation.

lib/bill_recognizer.rb
clemenshelm: It's a method on the `ImageProcessor`. It uses the RMagick `level` method: https://rmagick.github.io/image2.html#level

This basically increases the contrast on the bill.

lib/detectors/billing_period_detector.rb
tamacodechi: Perfect, thank you for clearing that up! 👍

lib/detectors/billing_period_detector.rb
tamacodechi: That makes sense, the system should be smart enough to calculate the correct BP regardless of the order in which they are collected 🙌

tamacodechi (approved):

PR #139, opened 2016-12-23
Report error when PDF cannot be read.

### LINKED ISSUE #137: Handle invalid pdfs gracefully
At the moment, PDFs which cannot be read lead to an exception like

```
NoMethodError: undefined method `page' for nil:NilClass
/myapp/lib/image_processor.rb:12:in `initialize'
```

Instead of failing with an exception, the recognizer should write an error back to the web app informing it about the problem.

Bills with invalid PDFs:
- [x] 5CCkGCCprokPBy2o6.pdf
- [x] CHhrbRhWvLofrWMDQ.pdf
- [x] 7vEgEHxN5cWQP8Tq7.pdf

clemenshelm:
Closes #137


spec/bill_recognizer_spec.rb
```diff
@@ -11,4 +11,12 @@
 
     expect(recognizer.recognize[:error]).to eq 'Unprocessable file type: .p7s'
   end
+
+  it 'reports when a bill cannot be read' do
+    # This spec will fail once Ghostscript supports this crappy PDF. Awesome!
```
tamacodechi, line 16: Great comment, because now when it does fail we will know why without having to dig around 👍

lib/image_processor.rb
```diff
@@ -2,13 +2,15 @@
...
 class ImageProcessor
+  class InvalidImage < StandardError; end
   include Magick
 
   def initialize(image_path)
     # Only read first page of bill
     image_path = "#{image_path}[0]"
     begin
       original_image = Image.read(image_path)[0]
+      raise InvalidImage, 'Cannot read image.' unless original_image
```
tamacodechi: Should the error be more descriptive? If this is in the case of PDF only then maybe something like "Cannot read image. Unsupported PDF." might be a little more descriptive 😀

tamacodechi (commented):
Awesome 😉  This isn't the case when the recognizer gives out that big block of text when leading a PDF, is it? (I ask because those still spit out recognized words, so I'm wondering if I've ever encountered an unsupported PDF)

tamacodechi (changes requested):
The new test in ```spec/bill_recognizer_spec.rb``` doesn't pass 🙈

lib/image_processor.rb
clemenshelm: It might be more descriptive, but it also might be wrong. If there is no image, it doesn't necessarily mean there's an error with the PDF. It could also be a JPG or a PNG in an invalid format. So I'd rather go with being more generic than with being more specific, but wrong 😉

clemenshelm:
No, the big block of text are probably just warnings. If the PDF is invalid, it won't be able to recognize anything because there won't be an image.

lib/image_processor.rb
tamacodechi: Ahh makes sense, I meant it more as an example where we extract the extension. So we can't really pass the file extension because it might simply be in the incorrect format? Then again the words "unsupported" kind of indicate that it is an unsupported file of that extension, and in the case of pngs and jpgs we know that it is that they're in an invalid format?

tamacodechi:
Yeah that I got I meant more that the spec isn't getting what it expected on Codeship. Odd, I fetched your branch and it all passed on my machine, so I would chalk it up to the environment on Codeship, right?

clemenshelm:
Just to clarify why there are so many commits and strange new files: I changed Codeship to use Docker to give us full control over the library versions we use (because there was a different version of Ghostscript used which made the test fail).

tamacodechi (approved):
I'd love to pick your brain on some of the changes but seeing as merging this is kinda urgent I'll approve first and ask questions later, because I don't see any changes that need to be made 👍

clemenshelm:
Ok, I merged it 😄 

Feel free to comment and I'll answer! 😉

PR #141, opened 2016-12-28
Make Unprocessable File Error pass recognizer version

### LINKED ISSUE #140: Pass version with recognition error
At the moment an error is passed to the web application simply as `{ error: 'some error' }`. Like the `recognitionResult`, this error object should also contain the current `recognizerVersion`.

tamacodechi:
Closes #140

~NOT READY YET.~ Okay it's ready now 😆


spec/bill_recognizer_spec.rb
```diff
@@ -5,10 +5,13 @@
...
-      file_basename: 'xcaEpkmTauDsZz9fk.p7s'
+      file_basename: 'xcaEpkmTauDsZz9fk.p7s',
+      recognizer_version: '13'
     )
     recognizer = BillRecognizer.new(retriever: retriever)
 
-    expect(recognizer.recognize[:error]).to eq 'Unprocessable file type: .p7s'
+    expect(
+      recognizer.recognize[:error]
+    ).to eq 'Unprocessable file type: .p7s. Recognizer version: 13'
```
clemenshelm: This is not how it should be. Instead, there should be an attribute `recognizerVersion`in the result, just like a normal result.

clemenshelm (changes requested):

spec/bill_recognizer_spec.rb
tamacodechi: Ahh makes sense! Thankfully should be an easy fix, so I'll get right on that 😉

clemenshelm (approved):

clemenshelm:
Yeees! Very nice now!

PR #142, opened 2017-01-03
Refactor to pass Rubocop standards

tamacodechi:
2 errors down, 49 to go! 😅 
With this PR there are no more troubles caused by the ```spec/spec_cache.rb``` file

Is this a good way to do it? I don't know if there are issues against passing around too many arguments from method to method.

### DIFF
--- a/spec/spec_cache.rb
+++ b/spec/spec_cache.rb
@@ -18,7 +18,14 @@ module SpecCache
   def cache_image(file_basename)
     bucket = 'chillbill-prod'
     region = 'eu-central-1'
-    image_path = cache_file(file_basename.to_s) do |path|
+    file_extension = File.extname file_basename.downcase
+
+    image_path = get_image_path(bucket, region, file_basename)
+    create_tempfile(image_path, file_extension)
+  end
+
+  def get_image_path(bucket, region, file_basename)
+    cache_file(file_basename.to_s) do |path|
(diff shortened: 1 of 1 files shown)


spec/spec_cache.rb
```diff
@@ -18,16 +18,24 @@ def cache_file(file_name)
   def cache_image(file_basename)
     bucket = 'chillbill-prod'
     region = 'eu-central-1'
-    image_path = cache_file(file_basename.to_s) do |path|
+    file_extension = File.extname file_basename.downcase
+
+    image_path = get_image_path(bucket, region, file_basename)
```
clemenshelm, line 23: The method name implies that the method just gets the image path when in fact it downloads and stores a file from S3.

I'd rather refactor it to something like

```ruby
image_path = cache_file(file_basename.to_s) do |path|
  download_from_s3(to: path)
end
```

clemenshelm: Btw, why do we need to call `file_basename.to_s`? What type is `file_basename`?

clemenshelm (changes requested):

tamacodechi:
I have no idea why but I have two branches doing very similar things for this solution. I'm going to close this PR and open a new one with the more correct solution. (This is what I hate about being away for so long, I had actually discovered this mess before I got sick 😞 )

PR #150, opened 2017-01-17
38 refactor bill image retriever

### LINKED ISSUE #38: lib/bill_image_retriever.rb - long method with too many ABCs
Rubocop still doesn't like our code, but these are fixes that require refactoring:

```
lib/bill_image_retriever.rb:15:3: C: Assignment Branch Condition size for save is too high. [21.56/15]
  def save
  ^^^
lib/bill_image_retriever.rb:15:3: C: Method has too many lines. [20/10]
  def save
  ^^^^^^^^
```

The save method does a lot that could be broken down into smaller methods, like determine the format of the bill that's being retrieved. We can refactor and compartmentalize in order to cut down on those ABCs and shorten the method 🎉

tamacodechi:
It may seem a bit silly to compartmentalize some of the things that are just a couple of lines, but if this method ever needs to do more it now has a couple of extra lines allowed to do so :)

I'm not 100% sure about about making image_file an instance variable but it's the only way without having to pass it around to every method, is this an okay way to do it? 

Closes #38


clemenshelm:
I'd rather do the following things:

- Deconstruct the URL in the constructor (also the file extension thing) and save the results as instance variables.
- Move the `Tempfile` part into your `get_bill_from_s3` method which should return the file
- Maybe then you don't even need the `determine_extension_validity` method anymore.

clemenshelm:
Why? Because the deconstructed URL is just another (more relevant) representation of the URL, so we won't need to store the URL in an instance variable anymore but only the relevant parts. This is also no heavy lifting, which I wouldn't put into a constructor (like downloading a file).

lib/bill_image_retriever.rb
```diff
@@ -18,26 +18,31 @@ class BillImageRetriever
...
-    logger.debug "bucket: #{bucket}, region: #{region}, key: #{key}"
+    logger.debug "bucket: #{@bucket}, region: #{@region}, key: #{@key}"
+
+    @file_extension = File.extname @key.downcase
+  end
 
-    file_extension = File.extname key.downcase
+  def save
+    create_temp_file_and_download_bill_from_s3
+    determine_extension_validity
```
clemenshelm: One thought here: I don't find it intuitive that `determine_extension_validity` returns the `image_file`. I think this method would read much more clearly something like this:

```ruby
def save
  ensure_extension_validity
  download_bill_from_s3
end
```
I’d do the extension check first because it's pointless to download the file when the extension is invalid.

clemenshelm (changes requested):

lib/bill_image_retriever.rb
tamacodechi: That makes loads of sense! 😅

tamacodechi: Hmm is it okay for ```download_bill_from_s3``` to return ```@image_file```? We need @image_file to be returned somewhere.

clemenshelm: I think yes, because to me it seams intuitive that `download_bill_from_s3` returns a file which has been downloaded. We probably won't need the `@image_file` instance variable anymore because it's only used within this method then.

lib/bill_image_retriever.rb
```diff
@@ -18,26 +18,32 @@ class BillImageRetriever
...
+  def download_bill_from_s3
+    @image_file = Tempfile.new ['bill', @file_extension]
+    s3 = Aws::S3::Client.new(region: @region)
+    s3.get_object(bucket: @bucket, key: @key, response_target: @image_file)
+    @image_file
+  end
 
-    case file_extension
+  def determine_extension_validity
+    case @file_extension
```
clemenshelm: This is a little bloated now. In fact it's enough to write

```ruby
# put this somewhere at the top of the class
VALID_EXTENSIONS = ['.pdf', '.png', '.jpg', '.jpeg'].freeze

def determine_extension_validity
  unless VALID_EXTENSIONS.include?(@file_extension)
    raise UnprocessableFileError, @file_extension
  end
end
```

clemenshelm (changes requested):

lib/bill_image_retriever.rb
tamacodechi: True, since we only have two cases and one of them only returns, this makes **so much more** sense 🎉  I'm a bit mad at myself for not catching that 😅

clemenshelm:
Is this ready for review now? Please let me know by tagging it with `needs_review` 😉

tamacodechi:
Ahh yes! Sorry I was waiting for the build to finish before that (I wasn't going to tag it if it was broken) and it took so long I just moved onto something else and forgot 😅

clemenshelm:
Oh, I see. There's a merge conflict now that I merged the other branch. Could you resolve it?

clemenshelm (approved):

clemenshelm:
Merge conflicts again 🦀

PR #153, opened 2017-01-18
Recognizer refactor interrim state

### LINKED ISSUE #39: lib/bill_recognizer.rb - long method with too many ABCs
As we all know, our recognize method is massive and therefore needs refactoring:

```
lib/bill_recognizer.rb:36:3: C: Assignment Branch Condition size for recognize is too high. [67.3/15]
  def recognize
  ^^^
lib/bill_recognizer.rb:36:3: C: Method has too many lines. [71/10]
  def recognize 
  ^^^^^^^^^^^^^
```

We could compartmentalize the detection and the calculations seperately, or maybe seperate by feature? Maybe we could all discuss on here what we think is best (that was some day in the future when we look at it and wonder why the heck we did it that way we'll also have documentation behind our line of thinking 😆 )

tamacodechi:
I think my eyes are so used to looking at this now that it will be hard for me to find things to move around anymore (except that we need to break things into smaller classes/modules/something).

I would love loads and loads of feedback if anyone has the time for it! Probably because I was so used to the recognizer the way it was, this feels very messy and like a lot of jumping around.

Closes #39 - for the most part, compartmentalization into classes will be a new issue.


lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
-      return {
-        error: e.to_s,
-        recognizerVersion: version
-      }
-    end
+    empty_database
+    version = fetch_recognizer_version
+    png_file = download_and_convert_image(version)
+
+    return png_file if png_file.is_a?(Hash)
```
tamacodechi: This is still the bad way it was. Rescuing outside of the method that does the downloading and converting (which are the two things that throw errors) stops the program without our desired result of rescuing the error and giving us the hash.

lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
+
+    image.apply_background('#fff')
+         .deskew
+         .normalize
+         .trim
+         .improve_level
+         .write_png!
+  end
 
+  def hocr(png_file)
```
tamacodechi: Is this a terrible name?

lib/bill_recognizer.rb
```diff
@@ -128,8 +152,58 @@ def recognize
...
     # }
-    price_words = PriceDetector.filter
-    logger.debug price_words.map { |word|
+  end
+
+  def configure_hocr(png_file)
+    tesseract_config = configure_tessarect
+    `tesseract "#{png_file.path}" stdout -l eng+deu #{tesseract_config}`
+      .force_encoding('UTF-8')
+    # logger.debug hocr
```
tamacodechi: I have never had this uncommented, can we remove it?

lib/bill_recognizer.rb
```diff
@@ -128,8 +152,58 @@ def recognize
...
+      top: adjusted_word[:top],
+      bottom: adjusted_word[:bottom]
+    )
+  end
+
+  def filter_words
+    DETECTORS.each(&:filter)
+  end
+
+  def log_price_words
```
tamacodechi: What do we use this for? We call it extra as its own thing from the recognize method.

lib/bill_recognizer.rb
```diff
@@ -138,85 +212,85 @@ def recognize
...
+      billingPeriod: calculate_billing_period,
+      currencyCode: calculate_currency,
+      dueDate: calculate_due_date,
+      iban: calculate_iban,
+      recognizerVersion: version
+    }
+  end
 
-    # image_file.close
+  def calculate_amounts(calculated_prices)
```
tamacodechi: How terribly confusing are all these price related method names? Personally haven't dealt with price detection. I'm pretty happy with how small each step is but I'm not sure if looking for the right place to change things won't be confusing because of the names.

lib/bill_recognizer.rb
```diff
@@ -138,85 +212,85 @@ def recognize
...
+  def calculate_prices
+    prices = PriceCalculation.new
+    {
+      net_amount: prices.net_amount,
+      vat_amount: prices.vat_amount
+    }
+  end
 
+  def calculate_totals(calculated_prices)
+    # Adapt recognition result to application schema
```
tamacodechi: Can I put this down as an issue an remove it? It's nice to have it there as an ugly reminder so we get sick of it and work on it too but so far it doesn't seem to have have baited anyone into working on it 😆

lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
-      }
-    end
+    empty_database
+    version = fetch_recognizer_version
+    png_file = download_and_convert_image(version)
+
+    return png_file if png_file.is_a?(Hash)
 
     # FileUtils.rm('./test.png')
     # FileUtils.cp(image_file.path, './test.png')
```
tamacodechi: Do we use these ever?

lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
-      Word.create(
-        text: word_node.text,
-        left: left,
-        right: right,
-        top: top,
-        bottom: bottom
-      )
-    end
+    extract_hocr_words(hocr_doc)
     # logger.debug Word.map {
```
tamacodechi: Since we have both the ```puts``` and the ```logger``` version, and we only ever use the ```puts``` version, can we remove this logger version and add it when we start using it? It's always been commented out both on master and on my end.

lib/calculations/relative_date_calculation.rb
```diff
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
 class RelativeDateCalculation
-  def initialize(words)
-    @words = words
-  end
-
   def relative_date(date_relative_to)
```
tamacodechi, line 3: I feel like there should be a default here... (irrelevant to recognizer but I found it along the way)

lib/calculations/vat_number_calculation.rb
```diff
@@ -2,13 +2,12 @@
 require_relative '../detectors/vat_number_detector'
 
 class VatNumberCalculation
-  def initialize(words, customer_vat_number:)
-    @words = words
+  def initialize(customer_vat_number:)
```
tamacodechi, line 5: When we get a bill with no customer vat (does this ever even happen?) do we just get a key with no value assigned? When writing up the test for the behaviour when there is no Vat Number detected it kept getting annoyed if I would pass anything without the keyword ```customer_vat_number``` so I just wanted to make sure that we always get this, even if it's for some reason empty.

spec/calculations/date_calculation_spec.rb
```diff
@@ -220,11 +210,8 @@
       bottom: 309
     )
 
-    due_date_calculation = DateCalculation.new(
-      DateTerm.dataset
-    ).due_date
-
-    expect(due_date_calculation).to eq DateTime.iso8601('2015-04-15')
+    due_date_calculation = DateCalculation.new
+    expect(due_date_calculation.due_date).to eq DateTime.iso8601('2015-04-15')
```
tamacodechi: Also irrelevant to the recognizer itself but for when I'm making the code more uniform, do we prefer calling the method within the expects or outside when we create the new Calculation?

eg.

```due_date_calculation = DateCalculation.new.duedate```
or
```expect(due_date_calculation.due_date).to eq```

File: spec/calculations/vat_number_calculation_spec.rb [hunk 12]
```diff
@@ -20,11 +20,15 @@
...
-      VatNumberTerm.dataset,
       customer_vat_number: 'ATU67760915'
     )
     vat_number_calculation.vat_number
 
     expect(vat_number_calculation.vat_number).to eq 'EU372001951'
   end
+
+  it 'returns nil if there is no vat number' do
+    vat_number = VatNumberCalculation.new(customer_vat_number: '')
```
tamacodechi: Only case where I'll need an empty customer_vat_number, but is this an okay way to do it? Just want to make sure.

lib/bill_recognizer.rb
clemenshelm: Why?

lib/bill_recognizer.rb
clemenshelm: Don't think so.

lib/bill_recognizer.rb
clemenshelm: I'd rather name this `recognize_words`.

lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
-
-      Word.create(
-        text: word_node.text,
-        left: left,
-        right: right,
-        top: top,
-        bottom: bottom
-      )
-    end
+    extract_hocr_words(hocr_doc)
```
clemenshelm: `create_words_from_hocr`?

lib/bill_recognizer.rb
clemenshelm: Yes

lib/bill_recognizer.rb
```diff
@@ -43,74 +43,98 @@
...
     ENV['TESSDATA_PREFIX'] = '.' # must be specified
-    tesseract_config = {
-      tessedit_create_hocr: 1,
-      tessedit_char_whitelist: %("#{Config[:tesseract_whitelist]}")
-    }.map { |k, v| "-c #{k}=#{v}" }.join(' ')
-    hocr =
-      `tesseract "#{png_file.path}" stdout -l eng+deu #{tesseract_config}`
-      .force_encoding('UTF-8')
-    # logger.debug hocr
+    hocr = configure_hocr(png_file)
```
clemenshelm: `configure` also might not be what's done here. Maybe `perform_ocr`?

lib/bill_recognizer.rb
clemenshelm: No idea. Let's remove it 😄

lib/bill_recognizer.rb
clemenshelm: I think this is ok.

lib/bill_recognizer.rb
```diff
@@ -138,85 +212,85 @@ def recognize
...
 
-    # image_file.close
+  def calculate_amounts(calculated_prices)
     amounts = []
+    return amounts if calculated_prices[:net_amount].nil?
+    totals = calculate_totals(calculated_prices)
+    amounts << {
+      total: (totals[:sub_total] + totals[:vat_total]).to_i,
+      vatRate: calculate_vat_rate(totals)
+    }
```
clemenshelm: I think this was much nicer the way it was before. Is it really necessary to move it into that many methods?

lib/bill_recognizer.rb
clemenshelm: Could you just do it right now? Doesn't seem to be a terrible amount of work 😉

lib/bill_recognizer.rb
```diff
@@ -138,85 +212,85 @@ def recognize
...
-        else
-          0
-        end }
-    end
+  def calculate_prices
+    prices = PriceCalculation.new
+    {
+      net_amount: prices.net_amount,
+      vat_amount: prices.vat_amount
+    }
```
clemenshelm: What's the benefit of putting it into a hash here?

lib/bill_recognizer.rb
```diff
@@ -138,85 +212,85 @@ def recognize
...
   end
 
-  private
+  def calculate_vat_rate(totals)
+    if totals[:sub_total].nonzero?
+      (totals[:vat_total] * 100 / totals[:sub_total]).round
+    else
+      0
+    end
+  end
```
clemenshelm, line 224: I think this is the only price-related method which really makes sense 😄

lib/calculations/relative_date_calculation.rb
clemenshelm: What would it be?

lib/calculations/vat_number_calculation.rb
clemenshelm: If there is none, the value is `nil` because this is the default value of `customer_vat_number` in the recognizer.

spec/calculations/date_calculation_spec.rb
clemenshelm: I don't really have an opinion on this 😄

spec/calculations/vat_number_calculation_spec.rb
clemenshelm: No, please pass `nil` in this case, not an empty String.

lib/calculations/vat_number_calculation.rb
tamacodechi: Good to know! 😉

lib/calculations/relative_date_calculation.rb
tamacodechi: I feel like it should be the invoice date but that would add complications/dependency and we won't ever really use it without passing a ```date_relative_to``` anyway, so I guess it's not absolutely necessary 😅

lib/bill_recognizer.rb
```diff
@@ -43,81 +43,96 @@
...
-      image_file = @retriever.save
-      png_file = preprocess(image_file.path)
+      png_file = download_and_convert_image
+
     rescue UnprocessableFileError, ImageProcessor::InvalidImage => e
       return {
         error: e.to_s,
         recognizerVersion: version
       }
     end
```
tamacodechi, line 92: Unfortunately this puts this methods 3 lines over. Maybe when we do further compartmentalization other method calls will be moved over, or is there any other way we can make our recognize method smaller? Or shall we just have to admit defeat and acknowledge that this is an important and therefore probably slightly bigger method? 😅

clemenshelm: Maybe putting the hash on one line? 😄 
Do we have a line length limit of 80? In the web app we have 100, I think we could also increase it here.

clemenshelm:
I think this looks very good now. Can we merge it soon? 😄

tamacodechi:
Sure, as soon as you're up for it! I already have the new issue (to deal with our new massive class) ready :D It will be nice to get this up and start getting comfortable with it!

PR #154, opened 2017-01-19
Refactor date calculation

tamacodechi:
It's a couple more lines but I feel like it's easier to read and it makes Rubocop happy 🎉  

Closes no issue, but makes out rubocop todo smaller.

### DIFF
--- a/lib/calculations/date_calculation.rb
+++ b/lib/calculations/date_calculation.rb
@@ -23,7 +23,7 @@ class DateCalculation
   end
 
   def find_labeled_invoice_date
-    return false if InvoiceDateLabelTerm.empty?
+    return nil if InvoiceDateLabelTerm.empty?
     DateTerm.right_after(InvoiceDateLabelTerm.first)
   end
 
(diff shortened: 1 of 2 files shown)


lib/calculations/date_calculation.rb
```diff
@@ -10,18 +10,23 @@ def initialize(words)
...
   end
 
+  def find_standalone_dates
+    DateTerm.all.select do |term|
+      term.started_periods.empty? && term.ended_periods.empty?
+    end
+  end
+
+  def find_labeled_invoice_date
+    return false if InvoiceDateLabelTerm.empty?
```
clemenshelm: I wouldn't return `false`. This way you either return a `DateTerm` or a boolean value which I find confusing. I'd rather return `nil` which is also falsy, but semantically means "there is no result".

clemenshelm (changes requested):

lib/calculations/date_calculation.rb
tamacodechi: Ahh oops, I forgot to change this back! Thank you!

clemenshelm (approved):

clemenshelm:
Sweet 🍭

PR #155, opened 2017-01-26
Date detector refactor

tamacodechi:
I imagine this will be the case with most detectors, but the find_x method is always way too long 😣 
I've made this PR for now, but I'd love some opinions on it before proceeding with it.


lib/detectors/date_detector.rb
```diff
@@ -11,69 +11,68 @@ class DateDetector
...
+
+    DateTerm.order(:first_word_id)
+  end
 
+  def self.find_short_period_and_hungarian_dates(words)
     result_words = find_dates(words, SHORT_PERIOD_DATE_REGEX, max_words: 2)
     hungarian_dates = find_dates(words, LONG_HUNGARIAN_DATE_REGEX, max_words: 1)
-    words -= (result_words + hungarian_dates)
+    result_words + hungarian_dates
+  end
```
tamacodechi: I understand why we remove the short period dates first, because otherwise period dates would get detected as either/or long period dates, but why are hungarian dates also part of the words we reduce from ```Word.all```? Do they also interfere with the long period dates?

If so would something like ```find_and_reduce_interfering_period_dates``` be a better name for this method?

lib/detectors/date_detector.rb
```diff
@@ -87,5 +86,13 @@ def find_dates(words, regex, after_each_word: nil, max_words: nil)
 
         affected_words
       end
```
tamacodechi, line 86: Does anyone else see a way to cut this block down? It's pretty much the same in most detectors and it would be great to cut this down. The fixes I can come up with are kind of fixes for the sake of fixes instead of it actually being better... I feel like moving most things out would defeat the purpose of the original ```find_x``` method

lib/detectors/date_detector.rb
clemenshelm: I think the year of the hungarian date could be considered in another date accidentally. But the easiest way to find out is to remove the `hungarian_dates` from the `reduced_words` and run the tests. If they still pass, there's no reason for the hungarian dates to be removed. 😉

lib/detectors/date_detector.rb
```diff
@@ -11,69 +11,68 @@ class DateDetector
...
+  end
 
+  def self.find_short_period_and_hungarian_dates(words)
     result_words = find_dates(words, SHORT_PERIOD_DATE_REGEX, max_words: 2)
     hungarian_dates = find_dates(words, LONG_HUNGARIAN_DATE_REGEX, max_words: 1)
-    words -= (result_words + hungarian_dates)
+    result_words + hungarian_dates
+  end
+
+  def self.find_long_dates_with_periods(words)
```
clemenshelm: Why *long* dates if the regex says `SHORT_PERIOD_DATE_REGEX`?

lib/detectors/date_detector.rb
clemenshelm: I think it belongs into a `Detector` superclass.

lib/detectors/date_detector.rb
```diff
@@ -11,69 +11,68 @@ class DateDetector
...
     private
 
       def find_dates(words, regex, after_each_word: nil, max_words: nil)
         affected_words = []
-        term = DateTerm.new(
-          regex: regex,
-          after_each_word: after_each_word,
-          max_words: max_words
-        )
+        term = create_new_term(regex, after_each_word, max_words)
```
clemenshelm: `create` is not a good term in this context as it implies that the term gets created in the database. Rather use `initialize`.

clemenshelm (changes requested):

lib/detectors/date_detector.rb
tamacodechi: Very true! I didn't think of it that way! 😅  I will try to keep that kind of context in mind for the future, thanks for the tip! 💃

clemenshelm (approved):

PR #156, opened 2017-01-26
44 refactor price calculation

### LINKED ISSUE #40: lib/calculations/price_calculation.rb - Non-local exit from iterator
On the issue of refactoring, we have this one `return` that rubocop hates. I tried replacing it but everything breaks. We can leave it on the rubocop todo list and it will get ignored by rubocop, but I feel that if we're "cheating" to get the result we want there must be something wrong with our code. Since it's not urgent we can also leave it on there for now, but maybe get around to a refactor? Thoughts?

```
lib/calculations/price_calculation.rb:43:11: W: Non-local exit from iterator, without return value. next, break, Array#find, Array#any?, etc. is preferred.
          return
          ^^^^^^
```

### LINKED ISSUE #44: lib/calculations/price_calculation.rb - long method with too many ABCs
There are too many ABCs and this method is too long so we could compartmentalize and shorten the method. Perhaps we can seperate the net and the vat? Or the calculation from the setting? Ideas?

```
lib/calculations/price_calculation.rb:29:3: C: Assignment Branch Condition size for net_and_vat is too high. [17.49/15]
  def net_and_vat
  ^^^
lib/calculations/price_calculation.rb:29:3: C: Method has too many lines. [14/10]
  def net_and_vat 
  ^^^^^^^^^^^^^^^
```

tamacodechi:
Closes #44 and closes #40 

Pleases Rubocop and is hopefully more legible 🐑


mina1604 (commented):
looks good :)

lib/calculations/price_calculation.rb
```diff
@@ -29,23 +29,30 @@ def calculate
...
+  def calculate_and_assign_net_and_vat(total_word, net)
+    calculate_possible_vats(net).each do |vat|
+      next unless net + vat == total_word.to_d
+      @vat_amount = vat
+      @net_amount = net
+    end
+  end
+
+  def calculate_possible_vats(net)
+    @remaining_prices.select do |price|
```
clemenshelm: Please don't use instance variables to pass state in the context of a single method call. They are meant for longer-lasting information across several method calls (like `@net_amount` and `@vat_amount`).

Instead, pass `remaining_prices` as an argument.

clemenshelm (changes requested):

lib/calculations/price_calculation.rb
tamacodechi: Gotcha 😉  Sorry for all this instance variable mess, will read up more in my free time to make sure I get a better idea of when to use them. It seemed wrong to have to pass ``` remaining_prices ``` down two layers 😅

clemenshelm: It's not wrong what you did and probably many programmers will use them like you do. My experience is just to avoid state wherever possible. And instance variables always create state because they live as long as the object does. This can lead to tricky problems which you can easily avoid by not using them unless there's a good reason to 😄

clemenshelm (approved):

PR #165, opened 2017-01-27
48 rewrite price spec test

### LINKED ISSUE #48: spec/detectors/price_detector_spec.rb - rewrite test format
There are a couple of tests written in this format. The others cause no problem because the comma and the numbers are separated by a whitespace. In this case we need them together because it's a word and it matches our expected result. 

```
spec/detectors/price_detector_spec.rb:29:28: W: Within %w/%W, quotes 
and ',' are unnecessary and may be unwanted in the resulting strings.
    create_following_words(%w(54, 00))
                           ^^^^^^^^^^
```

It's in the rubocop todo list so it won't come up or get rewritten by `rubocop -a`, but maybe we can rewrite it eventually.

tamacodechi:
Closes #48


spec/detectors/price_detector_spec.rb
```diff
@@ -26,7 +26,23 @@
...
+    )
+
+    create(
+      :word,
+      text: '00',
+      left: 0.42179319371727747,
+      right: 0.44960732984293195,
+      top: 0.23014586709886548,
+      bottom: 0.23871266496874277
+    )
```
clemenshelm: How is this an improvement? Isn't the easiest fix for the `PercentStringArray` rule just to rewrite the method call to `create_following_words(['54,', '00'])`?

clemenshelm (changes requested):

spec/detectors/price_detector_spec.rb
tamacodechi: Ahh right this works! I just thought it would be good to have the actual data that came from a test but I've changed it to this leaner format. 

Rubocop autocorrect just breaks the test because it keeps the same format but removes the comma, so no really testing what we want which was the two words, one of which has a comma 🤖 💦

clemenshelm (approved):

clemenshelm:
Awesome! The merge conflicts in `.rubocop_todo.yml` are a pain in the pass 😉

PR #166, opened 2017-01-27
157 refactor image processor

### LINKED ISSUE #157: lib/image_processor.rb - method too long
screen shot 2017-01-26 at 16 21 33

tamacodechi:
Closes #157 

I hope this is okay. I kind of like it because the ```page``` and ```min_dimension``` variables are only needed so we can calculate the density we want, so it's nice to have them grouped in their own method.  Also makes Rubocop happy 😊 

I set ```original_image``` as ```@original_image``` so we wouldn't have to pass it around. Since it gets destroyed (and I verified this as well just to make extra paranoid sure: ``` #<Magick::Image: (destroyed)>``` Is this okay, or is it better practice to just use a parameter since it's a function ie. we want what it's returning not just side effects?


lib/image_processor.rb
```diff
@@ -6,23 +6,26 @@ class InvalidImage < StandardError; end
...
-    density = 220_000.0 / min_dimension
     @image = Image.read(image_path) { self.density = density }[0]
   end
 
+  def calculate_density
+    page = @original_image.page
+    min_dimension = [page.width, page.height].min
+    # This will give us an image with at least 3000px on each dimension
+    220_000.0 / min_dimension
+  end
```
clemenshelm: I don't like the instance variable approach very much because the `original_image` is only used during initialization and isn't necessary in the further object lifecycle, but will obviously stick around, even if it's just a destroyed image.

My suggestion is to just pass the `page` as argument to the `calculate_density` function. What do you think about that?

clemenshelm (changes requested):

lib/image_processor.rb
tamacodechi: That's true, I still trying to get the hang of when to use which 😅 

Would it be okay to pass ``` original_image ``` as a regular variable, @clemenshelm ? This will still keep rubocop happy (```initialize``` will keep up to 10 lines). If it makes more sense to pass page though then we can just ingore this and add it to Rubocops' todo permanently

clemenshelm: Why does it take more lines to pass `original_image.page` to the method than to pass `original_image`?

tamacodechi: Ahh, no this works perfectly! In fact it was what I meant but I omitted the ```.page ``` and I thought you wanted me to define page and then pass it which would be an extra line just for that variable declaration, so it was just a misunderstanding! 😆

clemenshelm (approved):

clemenshelm:
Very great! Please resolve the conflicts then let's ship it!

PR #169, opened 2017-02-01
47 refactor logging

### LINKED ISSUE #47: lib/logging.rb - rewrite nested methods
Rubocop doesn't like our nested methods, and [neither should we](https://github.com/bbatsov/ruby-style-guide#no-nested-methods).

```
lib/logging.rb:16:7: W: Method definitions must not be nested. Use lambda instead.
      def logger 
      ^^^^^^^
```

tamacodechi:
Closes #47 

I'm a little unsure about this one because Rubocop requests a lambda instead of nested definitions we have. All is good when I define it using a lambda but rubocop doesn't like the "unused" declaration ``` logger = -> ``` so I call it directly underneath, but surely we're calling this logger somewhere else which is why it was defined as it was before, right?


clemenshelm:
I'm almost certain that this code doesn't work. What's the exact error that Rubocop gives you?

tamacodechi:
```
lib/logging.rb:16:7: W: Method definitions must not be nested. Use lambda instead.
      def logger ...
      ^^^^^^^^^^
```

tamacodechi:
To be honest I have no idea how our logger works. I will definitely be reading up on the Logger class this week in my free time, but if I'm still lost, would it be okay to spend like 10 mins pairing just going through how we've implemented it?

lib/logging.rb
```diff
@@ -13,9 +13,8 @@ def logger
   # Addition
   def self.included(base)
     class << base
-      def logger
-        Logging.logger
-      end
```
clemenshelm, line 18: Maybe changing the method definition to

```ruby
define_method(:logger) { Logging.logger }
```

will fix the Rubocop issue.

tamacodechi: Yes this is so much better and works beautifully. Thank you!! 👏  🎉

clemenshelm (approved):
Please resolve the conflicts, then we can merge it.

PR #175, opened 2017-02-03
171 short year detected

### LINKED ISSUE #171: Short year detected
In bill inHQfeSqLniJys2dk.PDF the year `203` is detected.

tamacodechi:
Closes #171 
The issue is actually an OCR issue. But this fixes the fact that an incorrect date gets detected that should not.

**Actual data:**
screen shot 2017-02-03 at 10 23 07

**What gets read:**
screen shot 2017-02-03 at 10 21 46

### DIFF
--- a/lib/detectors/date_detector.rb
+++ b/lib/detectors/date_detector.rb
@@ -12,7 +12,7 @@ class DateDetector
   SHORT_ENGLISH_DATE_REGEX = /((?:#{days})-(?:Oct)-\d{4}$)/
   LONG_SLASH_DATE_REGEX = %r{((?:#{days})/(?:#{months})/\d{4}$)}
   FULL_GERMAN_DATE_REGEX =
-    /(\d+\. (?:März|April|September|Oktober|Dezember) \d+)/
+    /(\d+\. (?:März|April|September|Oktober|Dezember) \d{4})/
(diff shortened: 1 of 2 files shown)


clemenshelm:
Why doesn't this close the issue? The issue already states that 203 is detected …

tamacodechi:
Ahh okay, well it doesn't make it so the correct date is detected only that the incorrect date is **not** detected 😅

clemenshelm:
That's a reasonable fix, though 😄

clemenshelm:
Even if it's not correct, at least it's less wrong 😉

clemenshelm:
Can you associate it with the issue and can I merge it then?

clemenshelm (approved):

tamacodechi:
True, to me it was more: the actual problem is that 2016 gets detected as 203 6, but this stops _another_ problem which is that 203 gets detected as a date, so it doesn't fix **this** issues, but it certainly fixes an issue the recognizer had 😁  So I didn't feel this qualified as closing the issue but rather fix a by product of the actual issue, where we should still need to fix that OCR issue along the way.

Awesome, will do 🎉 

 >The slightly obsessive person in me is now glad that we have data making us get rid of all those nasty ```d+``` regexes. Can't wait 'til they all need ```20\d{2}``` kind of limitation 😆

PR #178, opened 2017-02-08
148 correct bill orientation

### LINKED ISSUE #148: Correct bill orientation with EXIF entry
Some bills are displayed and processed sideways, especially when they are recorded with a mobile phone camera. In many cases, these images contain an EXIF entry indicating the orientation, though.

An example is bill gCQ76uE6qLYhEdsY9.JPG. When you open it with on your computer, it will get displayed correctly since the viewer app will consider the orientation EXIF entry. RMagick also gives us the possibility to use this tag: https://rmagick.github.io/image1.html#auto_orient_bang

Let's use it!

Also we can access this EXIF entry  in RMagick with [get_exif_by_entry](https://rmagick.github.io/image2.html#get_exif_by_entry). This entry is called `Orientation`. More info on this EXIF entry: https://beradrian.wordpress.com/2008/11/14/rotate-exif-images/
Please include the orientation in the recognition result in a meaningful format so we can also display the bill with corrected orientation to the reviewer.

tamacodechi:
Closes #148 
Questions on the comments :)
Hope this does the trick 🎉


lib/bill_recognizer.rb
```diff
@@ -118,8 +118,10 @@ def preprocess(image_path)
 
     @width = image.image_width
     @height = image.image_height
+    @orientation = image.get_orientation
```
tamacodechi: Again with the instance variables 😅  Does it make more sense in the case? We can only really ```get_orientation``` if we have ```image``` for which we need to be in this method or call everything all over again in it's own method.

lib/image_processor.rb
```diff
@@ -25,6 +25,16 @@ def calculate_density(page)
     220_000.0 / min_dimension
   end
 
+  def get_orientation
+    orientation = @image.get_exif_by_entry('Orientation')
+    orientation.first.last
```
tamacodechi: This ```.first.last``` is because ```get_exif_by_entry``` returns an array with an array, in this case: ```[["Orientation", "6"]]``` It's a bit weird but it does the job. I tried looking around but I didn't find any other kind of accessor, is it clear or should I leave a comment in the code in case anyone can't fathom why we do ```.first.last``` ? 🤔

lib/bill_recognizer.rb
```diff
@@ -116,15 +116,18 @@ def download_and_convert_image
   def preprocess(image_path)
     image = ImageProcessor.new(image_path)
 
-    @width = image.image_width
-    @height = image.image_height
+    @orientation = image.calculate_orientation
```
tamacodechi: Again with the instance variables 😅 Does it make more sense in the case? We can only really get_orientation if we have image for which we need to be in this method or call everything all over again in it's own method.

lib/bill_recognizer.rb
```diff
@@ -116,6 +116,9 @@ def download_and_convert_image
   def preprocess(image_path)
     image = ImageProcessor.new(image_path)
 
+    @orientation = image.calculate_orientation
```
tamacodechi: Again with the instance variables 😅 Does it make more sense in the case? We can only really get_orientation if we have image for which we need to be in this method or call everything all over again in it's own method.

tamacodechi:
It is now ready for review 👍

lib/bill_recognizer.rb
clemenshelm: It's ok, but I think you could improve the naming. Here you're talking about orientation, but at the bottom you're talking about `clockwiseRotationsRequired` all of a sudden 😉 

Maybe you could rename the variable and the method to also reflect that it's about rotations?

lib/image_processor.rb
```diff
@@ -4,7 +4,8 @@
 class ImageProcessor
   class InvalidImage < StandardError; end
   include Magick
-
+  attr_reader :image_width
+  attr_reader :image_height
```
clemenshelm: Better: `attr_reader :image_width, :image_height`

lib/image_processor.rb
```diff
@@ -25,6 +28,27 @@ def calculate_density(page)
...
+    case orientation.first.last
+    when '6'
+      1
+    when '3'
+      2
+    when '8'
+      3
+    else
+      0
+    end
```
clemenshelm: I'm not sure if Rubocop will like it, but it's possible to write simple case statements more concise:

```ruby
case orientation.first.last
when '6' then 1
when '3' then 2
when '8' then 3
else 0
end
```

spec/image_processor_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+require 'open-uri'
+require_relative '../lib/bill_image_retriever'
+require_relative '../lib/image_processor'
+
+describe ImageProcessor do
+  it 'auto corrects the orientation of a bill image' do
+    file_id = 'gCQ76uE6qLYhEdsY9'
+    file_url =
+      "https://chillbill-prod.s3-eu-central-1.amazonaws.com/#{file_id}.JPG"
```
clemenshelm: Instead of downloading images here, you could just create test images with RMagick: https://rmagick.github.io/image1.html#new (and save them)
Setting the EXIF data should also work: http://stackoverflow.com/questions/20901713/how-to-write-custom-metadata-to-an-image-created-by-rmagick

Then you don't have a dependency on the bill being there and on the internet connection. Plus you specify the dimensions explicitly which makes the test more understandable.

clemenshelm (changes requested):

spec/image_processor_spec.rb
tamacodechi: Ahh great idea! Yes, I hated that this partially depended on the retriever which is besides the point of these tests! Thank you, this is great 👏

lib/image_processor.rb
tamacodechi: Rubocop didn't mind but I do prefer this myself too, so I will change it ;)

lib/image_processor.rb
tamacodechi: Will do 👍

lib/bill_recognizer.rb
tamacodechi: Perfect! I wasn't sure if it was necessary. I also prefer it because it's more consistent but also then the variable name would be quite long and I didn't know if it was just my being nitpicky 😅

clemenshelm:
@Althaire I didn't figure out how to write the EXIF-Tag, probably it's not even possible with RMagick (the PNG stuff might have been other metadata because PNGs don't even support EXIF from what I've read).

Instead I created a 2x3 test image file with this tag and it works fine 😄

clemenshelm (approved):

PR #209, opened 2017-03-29
198 invoice number recognition

### LINKED ISSUE #198: invoice number recognition
The recognizer should start recognizing invoice numbers of frequent bills (like Billa etc.)

tamacodechi:
closes #198


lib/calculations/invoice_number_calculation.rb
```diff
@@ -0,0 +1,18 @@
...
+require_relative '../models/invoice_number_term'
+require_relative '../models/dimensionable'
+
+class InvoiceNumberCalculation
+  def invoice_number
+    return nil if InvoiceNumberTerm.empty?
+    if InvoiceNumberLabelTerm.any?
+      invoice_number =
+        InvoiceNumberTerm.right_after(InvoiceNumberLabelTerm.first) ||
+        InvoiceNumberTerm.right_below(InvoiceNumberLabelTerm.first)
```
clemenshelm: better:

```ruby
label = InvoiceNumberLabelTerm.first
if label
  invoice_number = InvoiceNumberTerm.right_after(label) || InvoiceNumberTerm.right_below(label)
…
```

lib/calculations/invoice_number_calculation.rb
```diff
@@ -0,0 +1,18 @@
...
+    else
+      invoice_number = InvoiceNumberTerm.where(needs_label: true).first
```
clemenshelm: I don't understand this. Why do you query for an `InvoiceNumberTerm` which needs a label when there is no label?

lib/detectors/invoice_number_label_detector.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+require_relative '../models/invoice_number_label_term'
+
+class InvoiceNumberLabelDetector
+  INVOICE_NUMBER_LABELS_REGEX = /(Re-Nr:|Bon-ID|Rech.Nr:|Beleg-nr.:)/
+  SINGLE_WORD_INVOICE_NUMBER_LABELS_REGEX =
+    /(Rechnungsnummer:|Rechnung:|Rechnungsnummer)/
+  MULTI_WORD_INVOICE_NUMBER_LABELS_REGEX =
+    /(Invoice number:|Billing ID|Rechnung Nr.:)/
```
clemenshelm: Oh, this is where you meant that it's kind of a random partition of the Regex 😄 

What I'm wondering: Since all of these Regexes are just whole strings, couldn't we just make one large array of strings up there and then merge them into one Regex? We do something similar already in the `CurrencyDetector`.

lib/models/dimensionable.rb
```diff
@@ -22,11 +22,18 @@ def right_before(current)
     def right_after(current)
       all.find do |following|
         (following.left > current.right) &&
-          (following.left - current.right) < (following.height * 19) &&
+          (following.left - current.right) < (following.height * 32) &&
```
clemenshelm, line 25: Could we solve this differently here? I think we talked about the problem that `height` is relative to the page height and therefore not really suitable for representing a space (because width and height are independent from each other). I think the best solution would be to define a "letter width" as `current.width / current.text.size` (or something like that) and then use a multiple of that as the threshold.

lib/models/word.rb
```diff
@@ -13,7 +13,7 @@ def next
   end
 
   def follows(previous_word)
-    max_space_width = previous_word.height
+    max_space_width = previous_word.height * 1.76
```
clemenshelm, line 16: Here it would also be nice to actually use the letter width mentioned above.

spec/calculations/invoice_number_calculation_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    invoice_number_calculation = InvoiceNumberCalculation.new
+
+    expect(
+      invoice_number_calculation.invoice_number
+    ).to eq '0547-20151202-02-5059'
+  end
+
+  it 'returns nil if there is no invoice number' do
+    invoice_number = InvoiceNumberCalculation.new
+    expect(invoice_number.invoice_number).to be_nil
```
clemenshelm: Bad naming (`invoice_number.invoice_number`)

clemenshelm (changes requested):
🍰

lib/models/dimensionable.rb
tamacodechi: I did built this but the multiplier was also very large because there are some cases where it's the word right after but the information is in columns eg.

| Invoice number: | 23123123123123123123123 |
|:---|---:|
| Invoice date:   |                 12.3.45 |

So I felt it didn't make sense to build this fix when it's not really fixing our problem. Should I go ahead and implement it anyway?

clemenshelm:
I updated the credentials on master. If you merge master into this branch, it should work.

lib/calculations/invoice_number_calculation.rb
```diff
@@ -6,12 +6,13 @@
...
+    label = InvoiceNumberLabelTerm.first
+    if label
       invoice_number =
-        InvoiceNumberTerm.right_after(InvoiceNumberLabelTerm.first) ||
-        InvoiceNumberTerm.right_below(InvoiceNumberLabelTerm.first)
+        InvoiceNumberTerm.right_after(label) ||
+        InvoiceNumberTerm.right_below(label)
     else
-      invoice_number = InvoiceNumberTerm.where(needs_label: true).first
+      invoice_number = InvoiceNumberTerm.where(needs_label: nil).first
```
clemenshelm: `needs_label: nil` doesn't make much sense semantically. How about `needs_label: false`?

clemenshelm (changes requested):

clemenshelm (approved):

PR #216, opened 2017-04-10
added dependable scale factor

### LINKED ISSUE #205: Great bill with accurate recognition... except the VAT ID number
Bill ` KsWubxamfEAwc7CfT.pdf` has everything detected correctly except the VAT ID, which is not recognized at all.

Tr4in:
Closes #205

### DIFF
--- a/spec/detectors/vat_number_detector_spec.rb
+++ b/spec/detectors/vat_number_detector_spec.rb
@@ -581,8 +581,193 @@ describe VatNumberDetector do
     expect(vat_numbers.map(&:to_s)).to eq ['DE147645058']
   end
 
-  # Many samples to show that the test fails
-  it 'detects nothing if median high is low' do
+  # Large sample size due to median height factor
+  it 'detects a VAT ID number using a larger median height' do
+    # from bill KsWubxamfEAwc7CfT.pdf
+    create(
+      :word,
+      text: '*',
+      left: 0.12238219895287958,
+      right: 0.12467277486910995,
+      top: 0.1343663274745606,
+      bottom: 0.13598519888991675
+    )
+
+    create(
+      :word,
+      text: 'E-Mail',
+      left: 0.649869109947644,
+      right: 0.6858638743455497,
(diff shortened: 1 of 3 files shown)


spec/detectors/vat_number_detector_spec.rb
```diff
@@ -580,4 +580,82 @@
     vat_numbers = VatNumberDetector.filter
     expect(vat_numbers.map(&:to_s)).to eq ['DE147645058']
   end
+
+  # Many samples to show that the test fails
+  it 'detects nothing if median high is low' do
+    create(
```
tamacodechi: Tiny nitpick, can you comment in the bill it's from?
The comment is also a little strange to understand. I think something like "Large sample size to calculate realistic median" would be better 😉

spec/detectors/vat_number_detector_spec.rb
```diff
@@ -580,4 +580,82 @@
...
+      :word,
+      text: 'ATU65315367',
+      left: 0.0824607329842932,
+      right: 0.20026178010471204,
+      top: 0.16281221091581868,
+      bottom: 0.17321924144310824
+    )
+
+    vat_numbers = VatNumberDetector.filter
+    expect(vat_numbers.map(&:to_s)).to be_empty
```
tamacodechi: Hmm... We don't want the test to expect it to be empty really.
I think the proper approach would be:

- Change the multiplier back to 1.15 from 1.5
- Write a test with the sample that makes the acceptance test fail (ie. The multiplier _should_ be 1.5 but for now in the code it's 1.15 so it fails)
- Add the ternary operator again so that the failing test passes

This would demonstrate why you need that `scale_factor`

The current way it is here, the detector is actually displaying behaviour we don't want, which is that it's not detecting that VAT ID.

tamacodechi (changes requested):
Awesome! 🙌

tamacodechi (approved):
Awesome! Great job! 👍

PR #218, opened 2017-04-11
fixed future dates

### LINKED ISSUE #206: Fix future dates
The bills `vGmK76dSSMrLQ8axN.pdf` and `rokPNRvjveNpX62jB.pdf` both have excellent recognition results however they both have an incorrect invoice date set in the future.

Tr4in:
Closes #206


lib/detectors/date_detector.rb
```diff
@@ -25,12 +25,12 @@ def self.filter
     reduced_words = filter_out_interefering_date_terms
     find_long_dates_with_periods(reduced_words)
 
+    find_multi_word_dates(reduced_words)
```
tamacodechi, line 28: Does the test you added need this change to pass? 😅 
If so, please change the name to reflect that. Thanks 👍

tamacodechi (changes requested):
Please autogenerate the rubocop todo :) `to_datetime` being one line longer is making rubocop angry 😉

spec/detectors/date_detector_spec.rb
```diff
@@ -324,6 +324,38 @@
...
+    )
+
+    create(
+      :word,
+      text: '2017',
+      left: 0.06833333333333333,
+      right: 0.09666666666666666,
+      top: 0.5585394581861013,
+      bottom: 0.5656065959952886
+    )
```
tamacodechi, line 354: Can you add a blank line after this parethesis? That way it looks just as nice as all our other tests 😂

spec/detectors/date_detector_spec.rb
```diff
@@ -712,6 +746,48 @@
...
+
+    create(
+      :word,
+      text: '51/5/38',
+      left: 0.5363219895287958,
+      right: 0.5765706806282722,
+      top: 0.005319148936170213,
+      bottom: 0.01156336725254394
+    )
```
tamacodechi, line 785: I don't think this should get detected at all 🤔
It's not a date.

tamacodechi (changes requested):

spec/detectors/date_detector_spec.rb
Tr4in: Although it is not a valid date it matches short slash regex

tamacodechi: You could limit the regex to accept dates from this decade for now :) I guess this solution won't work for all cases but it'll work for now

tamacodechi (changes requested):
Please update the rubocop todo ;)

spec/detectors/date_detector_spec.rb
clemenshelm: I'd prefer not to, the decade isn't that long anymore and I'd like to avoid that things stop all of a sudden on Jan 1, 2020 😉

clemenshelm: The fact that there's `51` and `38` disqualifies it as a date as only one of them can be a year. Please update the regex appropriately.

tamacodechi: I was thinking that too, and also there might be non dates that fall into this decade, so it's not a very healthy solution 😔

tamacodechi (approved):

PR #219, opened 2017-04-11
211 qr code detection

### LINKED ISSUE #211: Read QR Codes [Part 1]
- [ ] Integrate Zbar to identify the presence of QR codes in bills
- [ ] Add attribute of QR code presence to bill results 

_When we can query for at least ~50 bills with QR codes we can work on Read QR Codes [Part 2]._

tamacodechi:
Closes #211
🤖 🎉


lib/bill_recognizer.rb
```diff
@@ -117,17 +118,25 @@ def preprocess(image_path)
...
+                    .normalize
+                    .trim
+                    .improve_level
+                    .write_png!
+    begin
+      Qrio::Qr.load(png_file.path).qr.text
+      @qr_code_present = true
+    rescue NoMethodError
+      @qr_code_present = false
+    end
```
clemenshelm: Nice. I just don't think this should be here because it's not part of the preprocessing but rather something similar to the OCR.

clemenshelm (changes requested):

clemenshelm (approved):

PR #242, opened 2017-05-06
231 ml data generation

### LINKED ISSUE #231: Make ML data generation nice
Currently there's a prototype of machine learning data generation. It works somehow but it is horrible code and therefore hard to maintain and to change. Make it nice so we can extend and change things quickly.

clemenshelm:
Closes #231


lib/models/price_term.rb
```diff
@@ -16,6 +16,10 @@ def initialize(attrs)
     super
   end
 
+  def before_save
+    self.price = to_d
+  end
```
tamacodechi: This is probably going to be foolish of me, I've also pulled this branch to see how everything works together and I can't find where this is being called 😅 What it does makes perfect sense, but I haven't found where it gets called.

lib/image_processor.rb
```diff
@@ -73,7 +73,7 @@ def improve_level
   end
 
   def trim
-    @image.fuzz = '80%'
+    @image.fuzz = '70%'
```
tamacodechi, line 76: Might sound like overkill but might it be worth it to test these to show these values are what we need based on a certain/some certain bills? 

It sounds silly but I can ask you, "Why 70% instead of 80%?" and you can tell me "because it works better" but I still won't know _why_ it works better 😂

docker-compose.yml
```diff
@@ -45,6 +45,7 @@ tests:
   build: .
   volumes:
     - .:/myapp
+    - ./tmp/tmpdir:/tmp
```
tamacodechi, line 48: If I understand correctly this is just for when we run this service and mainly for the ML data?

Rakefile
```diff
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
+require_relative 'spec/spec_cache_retriever'
+require_relative 'lib/bill_recognizer'
+Dir["lib/tasks/machine_learning/*.rake"].each { |file| import file }
```
tamacodechi, line 4: 👍 Really cool! Should we move our other vanilla rake tasks into separate files and make a rake lib so it's a little more uniformly structured?

lib/boot.rb
```diff
@@ -16,6 +16,7 @@
 DB.create_table :price_terms do
   primary_key :id
   String :text
+  BigDecimal :price, size: [10, 2]
```
tamacodechi: What does size do, does it set the base?

lib/tasks/machine_learning/add_prices.rake
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+namespace :machine_learning do
+  desc 'Add recognized prices to imported bill data'
+  task :add_prices do
+    require 'yaml/store'
+    bill_files = Dir['data/bills/*.yml']
```
tamacodechi, line 6: Silly question, but say hypothetically someone wants to run these tasks locally, since the data is  not shared/versioned/*whatever the term is for being on the repo* where do we get it from? 🤔 My guess is we share the files on site...?

lib/tasks/machine_learning/add_prices.rake
```diff
@@ -0,0 +1,86 @@
...
+      store['amounts'].each do |amount|
+        vat_rate = amount['vatRate']
+
+        prices = extractor.extract_prices(for_amount: amount)
+
+        %i(total vat).each do |attr|
+          candidates = prices.send(attr)
+          price_key = "#{attr}_#{vat_rate}"
+          store["#{attr}_prices_candidates"][price_key] = candidates.map(&:to_h)
+          store["#{attr}_prices"][price_key] = nil
```
tamacodechi, line 41: Why do we need to set it to nil? Is it for the next total vat?

lib/tasks/machine_learning/generate_csvs.rake
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+namespace :machine_learning do
+  desc 'Generate CSV files for R from bill data'
+  task :generate_csvs do
+    require 'csv'
+    require 'securerandom'
+
+    def add_id(price)
+      require 'securerandom'
```
tamacodechi: Do we need to require it inside here as well as outside? 😮

lib/tasks/machine_learning/import_bill_data.rake
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+namespace :machine_learning do
```
tamacodechi, line 2: I haven't found any tasks that conflict in terms on namespace, should we start doing this for all rake tasks just for context/to avoid namespace problems in the future?

lib/tasks/machine_learning/generate_csvs.rake
```diff
@@ -0,0 +1,54 @@
...
+            vat_price = bill['vat_prices']["vat_#{vat_rate}"]
+            correct_tuples_csv << [
+              bill['_id'], amount_price['_id'], vat_price['_id'], vat_rate
+            ]
+          end
+        end
+      end
+    end
+  end
+end
```
tamacodechi, line 53: It's a long slow read to actually understand what's going on but I don't see any other way of doing it. The comments are great, thank you for adding those in! The logic is sound but it still wouldn't have much meaning unless you know exactly what's going on 👍

lib/tasks/machine_learning/import_bill_data.rake
```diff
@@ -0,0 +1,31 @@
...
+      store.transaction do
+        store['_id'] = bill[:_id]
+        store['image_url'] = bill[:imageUrl]
+        store['amounts'] = bill[:accountingRecord][:amounts].map(&:to_h)
+      end
+    end
+
+    puts "Successfully imported #{limit} bills."
+  end
+end
```
tamacodechi, line 31: 👍

lib/tasks/machine_learning/list_bills.rake
```diff
@@ -0,0 +1,27 @@
...
+      bill_label = "Bill #{bill['_id']}"
+      if problems.empty?
+        puts "👍🏼  #{bill_label}".green
+      else
+        puts "🙀  #{bill_label} has got the following problems:".red
+        problems.each { |p| puts "- #{p}".red }
+      end
+    end
+  end
+end
```
tamacodechi, line 27: Awesome ✌️ Sounds silly but the emojis are a great way to be able to skim to find problems :D

tamacodechi (approved):
Looks great! 🎉 This is some really cool work! I'm really excited how this will pave the way for working with other attributes as well

Rakefile
clemenshelm: Yeah, that's probably a good idea since they handle very different things. That's a separate chore task, though. 😄

docker-compose.yml
clemenshelm: I created this so tempfiles are accessible outside of the docker container, otherwise it's hard to inspect them for debugging. So the `/tmp` dir of the docker container will be available at `./tmp/tmpdir`.

clemenshelm: It might also be useful outside of the ML tasks.

clemenshelm: Especially for debugging the image processor functionality 😄

lib/boot.rb
clemenshelm: Number of total digits and number of decimal places.

lib/image_processor.rb
clemenshelm: It's the tolerance which we accept when trimming the bill. It was clearly too high because it trimmed across pretty dark grey logos 😄 

I'm pretty confident that this doesn't break anything else. If we encounter problems with it, it would be probably good to add some tests with example bills which make sure trimming and other operations are performed correctly.

lib/models/price_term.rb
clemenshelm: Yeah, true. I don't need it anymore after some refactoring. I'll remove this and also the database column.

lib/tasks/machine_learning/add_prices.rake
clemenshelm: They can generate the files by running `rake machine_learning:import_bill_data`. But I also created an S3 bucket which contains the already correctly processed bills. Probably I'll create another task in the future which will automatically sync the directory with S3.

lib/tasks/machine_learning/add_prices.rake
clemenshelm: Setting it to nil adds an entry to the YAML file without any value. That's what I want because this tasks creates some kind of scaffolding YAML file that still needs to be filled in manually. So  for example one of the `total_prices_candidates` needs to be copied to the `total_prices`.

lib/tasks/machine_learning/generate_csvs.rake
clemenshelm: No, of course not. Thanks 😄

lib/tasks/machine_learning/generate_csvs.rake
clemenshelm: Yes, I imagine it's particularly hard without having the YAML files at hand 😄

lib/tasks/machine_learning/import_bill_data.rake
clemenshelm: I think it's a good idea if we have a number of rake tasks on the same topic to group them under a common namespace. It's not so much about naming conflicts but about structure. Also when you run `rake -T`, these tasks appear grouped in the list.

PR #251, opened 2017-05-10
210 improve ocr

### LINKED ISSUE #70: Incorrect currency detected

### LINKED ISSUE #147: Irish vat number not detected
The vat number IE9692928F is not detected.

Bills:

- [x] GPtDdfL65GwhaP2sm.pdf
 - [x] gCjATsx9NFCgxssGh.pdf

### LINKED ISSUE #210: Improve OCR
- [x] Read up on tesseract and see what's available for us to improve on
- [x] Gather known OCR issues and identify possible solutions

Bills with OCR issues:
- [OCR Issues](https://github.com/clemenshelm/chillbill-recognizer/issues?q=is%3Aopen+is%3Aissue+label%3AOCR)
- local cache of OCR issue bills

tamacodechi:
✨IT✨IS✨DONE✨
Basically just upgrading to Tesseract 4.00 though. It's a bit hard to read because of the mess of commands but basically this installs all of the dependencies for Tesseract, clones and installs Leptonica (another dependency for Tesseract), and clones and installs the newest version of Tesseract.

_Fun problems along the way:_ Our linux version was "too old" because docker, clang++ was being tedious because of python, libjpeg8-dev is outdated, leptonica needed to be manually compiled as well because we needed 1.74 for tesseract 4.00, and _that_ was just a whole beast on it's own. I think it would be a nice blog post or read me because there were about 9 big problems and each of them required a lot of research to solve so it might help someone somewhere 😂 

Also fun, building the image now takes ~15 mins 😳

Closes #210 and closes #147 and closes #70

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,7 @@
 # Mostly taken from http://blog.giantswarm.io/getting-started-with-microservices-using-ruby-on-rails-and-docker/
 FROM ruby:2.3.1
 
-RUN apt-get update &&  apt-get install -y ghostscript g++ autoconf automake libtool autoconf-archive pkg-config libpng12-dev libjpeg62-turbo-dev libtiff5-dev zlib1g-dev libleptonica-dev && git clone https://github.com/DanBloomberg/leptonica.git && cd leptonica && mkdir m4 && autoreconf -vi && ./autobuild && ./configure && make && make install
-
-RUN git clone --depth 1 https://github.com/tesseract-ocr/tesseract.git && cd tesseract && ./autogen.sh && ./configure --enable-debug && LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make && make install && ldconfig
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Mostly taken from http://blog.giantswarm.io/getting-started-with-microservices-using-ruby-on-rails-and-docker/
 FROM ruby:2.3.1
 
-RUN apt-get update &&  apt-get install -y ghostscript tesseract-ocr tesseract-ocr-deu
+RUN apt-get update &&  apt-get install -y ghostscript g++ autoconf automake libtool autoconf-archive pkg-config libpng12-dev libjpeg62-turbo-dev libtiff5-dev zlib1g-dev libleptonica-dev && git clone https://github.com/DanBloomberg/leptonica.git && cd leptonica && mkdir m4 && autoreconf -vi && ./autobuild && ./configure && make && make install && git clone --depth 1 https://github.com/tesseract-ocr/tesseract.git && cd tesseract && ./autogen.sh && ./configure --enable-debug && LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make && make install && ldconfig
```
clemenshelm: Could we put all of these commands into separate `RUN` statements?

clemenshelm: Then we could also add an explaining comment to each of them 😄

tamacodechi: Kind of, I can separate the leptonica installation and the tesseract installation from the rest but I have to keep each of those in their own chunk because you can't really chain RUN statements (I'm not quite sure how to word if but if something from one RUN is dependent on another then you need to have them in the same one) Would it be enough to have 3? One for the general dependencies and tesseract dependencies, one for Leptonica and one for Tesseract?

clemenshelm: Yes, that would definitely be an improvement. Plus comments 😸

tamacodechi: Awesome, will do! 💃

PR #253, opened 2017-05-10
214 bill config dimensional ratio

### LINKED ISSUE #214: Make a bill config that is accessible by dimensionable
TL;DR: We need something that doesn't use the word height for calculating whitespaces in `right_after` and `follows` because we will need these enormous multipliers.

Maybe we create a table with the bill attributes that can be accessed by dimensionable to calculate the bill dimension ratio, or a yaml file...?

clemenshelm:
Closes #214


spec/detectors/billing_period_detector_spec.rb
```diff
@@ -129,6 +134,9 @@
   end
 
   it "it doesn't consider other dates as part of the billing period" do
+    # No idea what bill this comes from. Guessing dimensions ...
+    BillDimension.create_all(width: 3057, height: 4323)
+
```
tamacodechi, line 139: I will look for these (I think I have it noted somewhere, probably when I was writing the label detector) and label them and update these values! ✌️

spec/bill_dimension_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+require 'open-uri'
+require_relative '../lib/models/bill_dimension'
```
tamacodechi: Unless I am mistaken, you don't need this here because the spec helper loads it already 🎉

lib/models/bill_dimension.rb
```diff
@@ -0,0 +1,21 @@
...
+    def create_all(width:, height:)
+      BillDimension.create(name: 'bill_width', dimension: width)
+      BillDimension.create(name: 'bill_height', dimension: height)
+    end
+
+    %w(bill_width bill_height).each do |dimension_name|
+      define_method(dimension_name) do
+        BillDimension.find(name: dimension_name).dimension
+      end
+    end
```
tamacodechi, line 15: Is all of this easier than just making a bill dimensions table with both attributes and just accessing them via their label? 🐙

lib/models/dimensionable.rb
```diff
@@ -21,8 +21,9 @@ def right_before(current)
 
     def right_after(current)
       all.find do |following|
+        space_width = following.height * BillDimension.bill_ratio
```
tamacodechi, line 24: ✌️

tamacodechi (changes requested):
It's great! I'm a bit disappointed it doesn't fix that horrible multiplier we need for right_after, but of course it wouldn't be that simple 😅 We need to create follow up issues for using the ratio for `.follows`, `.right_below` and `.below` 😉

lib/models/bill_dimension.rb
tamacodechi: So that we can add more things in the future like trim width, etc...

tamacodechi (approved):

PR #259, opened 2017-05-15
Fix qr detection.

### LINKED ISSUE #236: QR detection very slow and crashes Merlin
The current QR code detection is very slow, causing many bills not to be detected because their detection job times out (> 200 seconds) and all other bills to be processed very slowly.

Also Merlin crashes about once a day due to memory issues which were very likely introduced with the QR code detection.

This is mainly because the QR code detector analyzes the whole page. Identify only relevant areas which could include a QR code and feed only these to the QR code. QRio does this already, but in Ruby.

Alternatively make zbar work. Probably this is even the better solution since zbar is a well-adopted library.

clemenshelm:
Closes #236


Dockerfile
```diff
@@ -2,13 +2,40 @@
...
 
 # This downloads and compiles Tesseract 4.00
-RUN git clone --depth 1 https://github.com/tesseract-ocr/tesseract.git && cd tesseract && ./autogen.sh && ./configure --enable-debug && LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make && make install && ldconfig
+RUN git clone --depth 1 --branch 4.00.00alpha https://github.com/tesseract-ocr/tesseract.git && \
+  cd tesseract && \
+  ./autogen.sh && \
+  ./configure --enable-debug && \
+  LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make && \
+  make install && \
+  ldconfig
```
tamacodechi, line 38: This all looks so much nicer 👏 👏 Good to know!

lib/qr_decoder.rb
```diff
@@ -0,0 +1,23 @@
...
+require 'zbar'
+
+class QRDecoder
+  include Magick
+
+  def initialize(image)
+    @image = image
+  end
+
+  def qr_code?
```
tamacodechi, line 11: It's clear because of `decoder` being defined directly above in the recognizer, but might it not be better to name this method slightly differently? Calling it from the other `qr_code?` makes it look kind of like a recursion.

spec/acceptance/bill_recognition_spec.rb
```diff
@@ -18,4 +18,13 @@
...
   end
+
+  it 'recognizes a bill with QR code' do
+    retriever = SpecCacheRetriever.new(file_basename: '3T2PCjhm5ovF4Mdce.pdf')
+    recognizer = BillRecognizer.new(retriever: retriever)
+
+    bill_attributes = recognizer.recognize
+
+    expect(bill_attributes[:qrCodePresent]).to be true
+  end
```
tamacodechi, line 29: Do we have any bills that have all the attributes nicely there _and_ a QR code? It would be nice to just have it all in one 🎉

Just wondering from our local caches, not that we should dig around before the QR code detection is fixed.

tamacodechi (changes requested):
Awesome 👍 This definitely looks like it'll work so much better than `qrio`. For the test files, can't we pull ones from our processed bills instead of dragging those images around on the repo?

lib/qr_decoder.rb
clemenshelm: Hmmm … I don't think so. In fact both methods have the same purpose: Returning whether there's a qr code or not. So to me this is why it makes sense to name them the same way.

Dockerfile
clemenshelm: Yeah, I got it from the Tesseract Dockerfile 😉

spec/acceptance/bill_recognition_spec.rb
clemenshelm: I think the purpose of this test is only to prove that the QR detection works. Adding the other attributes doesn't have much benefit since we already test this functionality above and makes the test harder to comprehend.

clemenshelm:
Regarding the images: I didn't find any, these were local ones that I recorded 😄 

But I also thought for the unit tests this makes more sense because we don't need all the other overhead (like downloading and preprocessing the image). For the bill orientation specs we also did it this way.

lib/qr_decoder.rb
tamacodechi: True... But also if two methods have the same purpose, aren't they a bit redundant? Because although it looks less like the other attributes, for that matter we could just have `qrCodePresent: QRDecoder.new(@image.image).qr_code?` because it's kinda of redundant to have two methods for returning the same value.

If `decoder` was not defined directly above it, it would very much look like a recursion because you would know what class `decoder` is. I think with fresh eyes it's a little confusing.

tamacodechi:
True, I forgot about those specs! 👍

lib/qr_decoder.rb
clemenshelm: Yes, but the point is that `decoder` *is* defined directly above it 😉 

I'm gonna change it to your version, though. Even less lines of code and not really less readable.

tamacodechi (approved):

PR #262, opened 2017-05-17
254 parent docker image

### LINKED ISSUE #254: Put packages in parent docker image
At the moment packages are compiled and installed when building a docker image. This means especially with every build on Codeship. Since we compile our own Tesseract version now, the build got very slow (from 4 minutes to 15 minutes) and it's unnecessary to compile them every single time.

Make a parent image with all compiled dependencies and store it in a repository so we can just retrieve it when we build the recognizer image.

tamacodechi:
It works 😭 😭 😭
... I was going a little crazy. I fell down the documentation rabbit hole.

The rake task also needs the deployment environment unfortunately. The task is not actually necessary because every developer can log in, pull the image and build the recognizer image. It's there for if someone wants it and in-tandem documentation for how to build the image. The nice thing is that once you've pulled the image, since it won't change very often, this will be super fast 🏃‍♀️ 

Please feel free to frown upon the sparkle emoji 😂 

Closes #254


tamacodechi:
screen shot 2017-05-17 at 15 10 31

_-throws laptop out the window-_
 😂 😂 😂 

@clemenshelm  you just have to give permission to Codeship's IAM AWS user, if I'm not mistaken ;)

clemenshelm:
Cool, good job! 🕺 

Things left to do:

- [x] Fix Codeship build
- [x] Make sure deployment still works with this.

tamacodechi:
Done 🎉 
I watched the AWS metrics to make sure it didn't kill anything and everything seems to be working well :) The metrics seem to indicate that the recognizer wasn't doing anything at times when there bills coming in, but it seems active once again now 🤔

PR #307, opened 2017-07-06
Clean up qr decoder and filter out only qr codes

### LINKED ISSUE #301: Make qr_code attribute contain ONLY QR codes, no other symbols
Use zbar to check for a symbol and then use the 
 symbology attribute from that symbol to filter out only qr codes.

tamacodechi:
Closes #301

If we redeploy with this, getting test bills for the QR code parsing will be a lot easier when the bills get reprocessed 🎈

### DIFF
--- a/lib/qr_decoder.rb
+++ b/lib/qr_decoder.rb
@@ -9,13 +9,6 @@ class QRDecoder
   end
 
   def qr_code?
-    return false unless symbol?
-    symbology == 'QR-Code'
-  end
-
-  private
-
-  def symbol?
     tmp_image = @image.dup
     image_blob = tmp_image.to_blob do
       self.depth = 8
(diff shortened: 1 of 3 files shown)


lib/qr_decoder.rb
```diff
@@ -9,15 +9,25 @@ def initialize(image)
...
+  def symbol?
     tmp_image = @image.dup
     image_blob = tmp_image.to_blob do
       self.depth = 8
       self.format = 'PGM'
     end
     tmp_image.destroy!
 
-    zbar_image = ZBar::Image.from_pgm(image_blob)
+    @zbar_image = ZBar::Image.from_pgm(image_blob).process
```
clemenshelm: Couldn't we just write `ZBar::Image.from_pgm(image_blob).process(symbology: : qrcode)`?

clemenshelm (changes requested):
Nice! Please add tests. 😉

clemenshelm (approved):

PR #314, opened 2017-07-17
267 parse qr codes

### LINKED ISSUE #267: Parse QR codes
Now that we have enough QR code sample bills, parse the QR code and extract the bill attributes from it.

tamacodechi:
Closes #267 

I'm not terribly happy with the way it's implemented in the recognizer but it works 😅 I'd love some input on how to make that a bit nicer...

Also I've kept the uncompacted prices, is that something we would want? 🦄


lib/bill_recognizer.rb
```diff
@@ -192,14 +193,29 @@ def filter_words
...
+                    end,
       vatNumber: calculate_vat_number,
       billingPeriod: calculate_billing_period,
       currencyCode: calculate_currency,
-      dueDate: calculate_due_date,
+      dueDate:  if @qr_data
+                  @qr_data[:dueDate]&.strftime('%Y-%m-%d')
+                else
+                  calculate_due_date
+                end,
```
clemenshelm: I think it would be more readable to write something like

```ruby
# Calculate all things that need to get calculated anyway
attributes = {
 billingPeriod: calculate_billing_period,
  …
}
if @qr_data
  attributes.merge(@qr_data) # @qrdata should contain everything in the right format (especially dates)
else
  attributes.merge(
    amounts: calculate_amounts,
    …
  )
end
```

lib/qr_decoder.rb
```diff
@@ -16,7 +23,29 @@ def qr_code?
...
 
-    qr_codes = ZBar::Image.from_pgm(image_blob).process(symbology: :qrcode)
-    !qr_codes.empty?
+    @qr_codes = ZBar::Image.from_pgm(image_blob).process(symbology: :qrcode)
+    !@qr_codes.empty?
+  end
+
+  def decode_qr_code
+    return unless qr_code?
+    all_data = @qr_codes.last.instance_variable_get(:@data).split('_')
```
clemenshelm: You cannot rely on `@qr_codes` being set here. You silently require `qr_code?` to be called first.

Better solution: Make a private method `qr_codes` and use it instead of the instance variables:

```ruby
def qr_code?
  !qr_codes.empty?
end

def decode_qr_code
  return unless qr_code?
  all_data = qr_codes.last.data.split('_')
  …
end

private

def qr_codes
  return @qr_codes if @qr_codes

  tmp_image = @image.dup
  image_blob = tmp_image.to_blob do
    self.depth = 8
    self.format = 'PGM'
  end
  @qr_codes = ZBar::Image.from_pgm(image_blob).process(symbology: :qrcode)
end
```

clemenshelm: Why did you use `.instance_variable_get(:@data)`? Doesn't just `.data` work?

lib/qr_decoder.rb
```diff
@@ -16,7 +23,29 @@ def qr_code?
...
+    prices = all_data.find_all { |price| /\d,\d{2}/ =~ price }
```
clemenshelm:
```ruby
prices = all_data.grep(/\d,\d{2}/)
```

lib/qr_decoder.rb
```diff
@@ -16,7 +23,29 @@ def qr_code?
...
+    prices_and_vats = VAT_RATES.zip(prices).to_h
+    compacted_prices_and_vats = prices_and_vats.select do |_vat, price|
+      /^(?!.*0,00).*$*\d,\d{2}/ =~ price
+    end
+    processed_data = {
+      invoiceDate: DateTime.strptime(all_data[4], '%Y-%m-%d'),
+      dueDate: DateTime.strptime(all_data[4], '%Y-%m-%d'),
```
clemenshelm: This should already be in the right date format (as a string), so we can just merge it into the bill result.

lib/qr_decoder.rb
```diff
@@ -16,7 +23,29 @@ def qr_code?
...
+      amounts: [{
+        total: (BigDecimal.new(
+          compacted_prices_and_vats.values.first.sub!(',', '.')
+        ) * 100
+               ).to_i,
+        vatRate: compacted_prices_and_vats.keys.first
+      }]
+    }
+    processed_data
```
clemenshelm: You don't need the `processed_data` variable as the last statement (the result hash) is returned implicitly anyway.

clemenshelm (changes requested):
Nice! Looking forward to shipping this 😄

lib/bill_recognizer.rb
clemenshelm: The comments are only meant for you and don't have to go into the code 😄

tamacodechi: Hmm, I thought about this, but does it make sense to go through calculating things that we won't need anyway? Also if something depended on another value, like it often does with the dates, if our recognizer picked up something wrong it would carry over onto the other calculations and then the original value will be correctly overwritten. 

Those were my concerns 😰

lib/qr_decoder.rb
tamacodechi: Okay! That sounds good! I did want to make it a private method but I wasn't sure if overcomplicating it by splitting it up would be well received 😅

lib/qr_decoder.rb
tamacodechi: Whoops, left over from prying 😅  Thank you for pointing it out!! 😂

lib/qr_decoder.rb
tamacodechi: 👍

lib/bill_recognizer.rb
clemenshelm: This code doesn't go through calculating things that we won't need anyway. All these things are in the `else` branch.

I don't understand the second point entirely. Can you give me an example?

lib/qr_decoder.rb
clemenshelm: In general supposing your public methods to be called in a specific order is bad practice as it will break easily.

lib/bill_recognizer.rb
tamacodechi: Ahh okay I misunderstood and thought it would be overwriting the already recognized and calculated values 😅 The second point is moot with that clarification too 😂

lib/qr_decoder.rb
```diff
@@ -3,20 +3,60 @@
...
+        all_data[4], '%Y-%m-%d'
+      ).strftime('%Y-%m-%d'),
+      amounts: [{
+        total: (BigDecimal.new(
+          compacted_prices_and_vats.values.first.sub!(',', '.')
+        ) * 100
+               ).to_i,
+        vatRate: compacted_prices_and_vats.keys.first
+      }]
+    }
```
clemenshelm: A bit hard to read this way. Could you put the results into variables first and add them to the hash afterwards? A rule of thumb could be: if it fits into one line, put it into the hash, otherwise into a separate step 🙂

clemenshelm (changes requested):
Good job! Just one little remark about formatting.

clemenshelm (approved):
Great work! 🎉

PR #315, opened 2017-07-26
Remove trimming, create text box and rewrite test data now that trimming is removed

### LINKED ISSUE #310: Don't trim bills, use text boundaries instead
Trimming bills doesn't work reliably: Sometimes the bill is not trimmed sufficiently because of single black pixels, sometimes too much is trimmed (logos, light fonts).

Instead of trimming we want to use two boxes instead:

- The total width and height (bill dimensions) and
- the text box

The text box is the bounding box of all words of a bill page, so the minimum left and top and the maximum right and bottom coordinate of all words.

Relative word coordinates should be calculated from the bill dimensions, not from the text box (probably requires no changes). We can still calculate coordinates relative to the text box  later on if we need them.

Pass both the bill dimensions and the text box to machine learning.

tamacodechi:
Closes #310 
This is just the text box creation, testing and removal of trimming 🦊 ✨ 
I'm rewriting the test data now but I didn't want to clutter the PR with all of that yet.


lib/models/bill_dimension.rb
```diff
@@ -17,5 +19,14 @@ def create_all(width:, height:)
     def bill_ratio
       bill_width / bill_height
     end
+
+    def text_box_boundaries(
+      text_box_top:, text_box_bottom:, text_box_left:, text_box_right:
```
clemenshelm: Since the method name already informs about the scope "text_box", I think it's sufficient to name the arguments `top:, bottom:, left:, :right`.

lib/models/bill_dimension.rb
```diff
@@ -17,5 +19,14 @@ def create_all(width:, height:)
...
+    )
+      BillDimension.create(name: 'text_box_top', dimension: text_box_top)
+      BillDimension.create(name: 'text_box_bottom', dimension: text_box_bottom)
+      BillDimension.create(name: 'text_box_left', dimension: text_box_left)
+      BillDimension.create(name: 'text_box_right', dimension: text_box_right)
```
clemenshelm: Bill dimension names are good the way they are, though. Don't change them 😄

lib/bill_recognizer.rb
```diff
@@ -188,6 +189,20 @@ def create_word(word_node, adjusted_word)
     )
   end
 
+  def calculate_text_box
+    text_box_top = Word.select_order_map(:top).first
```
clemenshelm: Where does `Word.select_order_map` come from? It's neither in the pull request nor on master. Either way, you should come up with a more descriptive name of what's the purpose of this method. It does something like determining the maximum boundary of one dimension, right?

clemenshelm (changes requested):
Looking very good so far!

lib/bill_recognizer.rb
tamacodechi: It's a sequel querying method. Should I create a method that just does this in order to make it more legible? Although... within that method the reader might still wonder what the method does 🙈

lib/models/bill_dimension.rb
tamacodechi: Okay! 😀  So here I should keep them as `text_top` etc.. or just top, bottom, left, right?

lib/models/bill_dimension.rb
tamacodechi: Phew, okay! That's better, since it gets a bit long it would be nice to have it shorter as well 😊

lib/tasks/machine_learning/add_dimensions.rake
```diff
@@ -26,7 +26,12 @@ namespace :machine_learning do
       recognizer.download_and_convert_image
       store['dimensions'] = {
         'width' => BillDimension.bill_width,
-        'height' => BillDimension.bill_height
+        'height' => BillDimension.bill_height,
+        'text_box_top' => BillDimension.text_box_top,
+        'text_box_bottom' => BillDimension.text_box_bottom,
+        'text_box_left' => BillDimension.text_box_left,
+        'text_box_right' => BillDimension.text_box_right,
+        'bill_type' => 'A4'
```
clemenshelm: Wouldn't `format` be a nicer name in the context of dimensions?

lib/models/bill_dimension.rb
```diff
@@ -17,5 +19,12 @@ def create_all(width:, height:)
     def bill_ratio
       bill_width / bill_height
     end
+
+    def text_box_boundaries(top:, bottom:, left:, right:)
```
clemenshelm: Maybe `create_text_boundaries`? Otherwise it's not clear what will happen in this method.

lib/models/bill_dimension.rb
```diff
@@ -8,7 +8,9 @@ def create_all(width:, height:)
       BillDimension.create(name: 'bill_height', dimension: height)
```
clemenshelm, line 8: In a way it's strange that this method is called `create_all` but doesn't create the text boundaries. Why did you choose to make a separate method?

clemenshelm (commented):
Just a few questions about naming 🙂

lib/tasks/machine_learning/add_dimensions.rake
tamacodechi: Sure, just checked with @Thomas1e and that name works for him too ;)

lib/models/bill_dimension.rb
tamacodechi: By design we use the bill width and height for the words (which comes before finding the text box, as that depends on the found words). Also Rubocop hates us using more that 5 parameters... but I mean we could have them as one 🤷‍♀️

lib/models/bill_dimension.rb
tamacodechi: True 🤔 Will change it right away!

clemenshelm (approved):

PR #324, opened 2017-07-31
Support unknown QR codes.

### LINKED ISSUE #323: Exception with different QR codes
If a bill contains a QR code in an unsupported format (no cash register QR code), there is an exception.

clemenshelm:
Closes #323


clemenshelm:
@Althaire I'm already merging this because this exception happens very frequently and it's blocking Merlin.

PR #326, opened 2017-07-31
Fix error with large invalid dates.

### LINKED ISSUE #325: Exception when parsing dates
There is an exception that happens when very large numbers get parsed as dates.

clemenshelm:
Closes #325


clemenshelm:
@Althaire I'm already merging this because this exception happens very frequently and it's blocking Merlin.

PR #334, opened 2017-08-09
Expand QR code decoder to save more than one price

### LINKED ISSUE #309: Extend QR code parsing for multiple prices on one bill
So far we don't have enough, or rather any, test data to write this feature but there are bills that contain multiple prices! ⛱  The QR code decoder should be able to return multiple prices in an array.

tamacodechi:
Closes #309 

Question though, should the images of the QR codes  just be the QR codes themselves or is it alright as the bill? Probably need @Tr4in 's input on this one, right? 😉


clemenshelm (approved):
Excellent! 🥇

Issue #332, opened 2017-08-08
Wrong text-box dimensions


Thomas1e:
The dimensions of the text-box are sometimes wrong. Maybe there are some "artifacts" recognized as text. 

Example bill: 
- 4gf87ndsAgDhTjZmg.pdf  (text_box_top = 0, which is incorrect)
- QcgqQs8d5bkLqs65q.pdf (text_box_bottom > 1!!)
- iLWtAEvMEtQw2bauv.pdf (text_box_left = 0!, should be around 0.05)
- T62xs4gCJ4WtCZrbh.pdf (text_box_right > 1!)
- bRrcBNr9ma9RnXoFT.pdf (text_box_bottom >1, text_box_right >1)

clemenshelm:
Thanks for reporting! Just delete the test data on your machine if this happens and add the example here.

tamacodechi:
Another good example that could help us filter out the "artefacts" is `iEXgdeSEdWnovb6j6.pdf`

When we inspect the height and width of all the present words, there are many very very small ones that may be artefacts, or may be single letters. We can definitely investigate and use this to filter them out.

PR #336, opened 2017-08-10
289 weight recognized as price

### LINKED ISSUE #289: Weight recognized as price
In bill WmcA2uThGP5QaaciP.pdf, the weight (123,00) is recognized as price twice.

tamacodechi:
Closes #289


lib/models/price_term.rb
```diff
@@ -9,7 +9,7 @@ class PriceTerm < Sequel::Model
   include Dimensionable
 
   def valid?
-    super && Word.right_after(self)&.text != 'kg'
+    super && Word.right_after(self)&.text != 'kg' && Word.right_below(self)&.text != 'kg'
```
tamacodechi: Do you know a nicer way to put this rather than `x != z && y != z` ? To kind of say "neither x nor y equals z"
(I want a nor without having to extend the boolean classes 😅 )

clemenshelm: You could write `![Word.right_after(self), Word.right_below(self)].map(&:text).include?('kg')`, but I doubt that this would be more readable 😄

tamacodechi: 😂 

Yeah this is why I asked for feedback, the way it is is veeeery clear as to what's happening, but I thought there might be some wonderful ruby magic I might have forgotten

clemenshelm: You could also transform it with de Morgan's rule to `!(Word.right_after(self)&.text == 'kg' || Word.right_below(self)&.text == 'kg')` but yeah … I think it's ok the way it is.

clemenshelm: How does this actually make the test pass? The "kg" in the example is right above the term, not right below.

clemenshelm (changes requested):
Please clarify the conditions.

tamacodechi:
🙀 our `right_below` method was so lenient that it would allow things well above to get picked up as well!

Here is the new and shiny `right_above` method 🎉 🎉  I suspect it will soon be too lenient as well but as soon as I apply any limitation (say, the thing above can only be one space above) the test does not pass. Should I limit it with a despotic multiplier which works for this example?

spec/models/dimensionable_spec.rb
```diff
@@ -325,5 +325,29 @@
...
+        text: '123,00',
+        left: 0.7081151832460733,
+        right: 0.7653795811518325,
+        top: 0.41463979615473706,
+        bottom: 0.4232105628908965
+      )
+
+      result = Word.right_below(below)
+      expect(result).to be_nil
+    end
```
clemenshelm, line 365: Spec for right_above is missing.

tamacodechi: Ahh in dimensionable! Oops! 🙊 Thank you!

clemenshelm:
Have you checked that this solution actually solves the problem of the bill? I'm not so sure about it.

tamacodechi:
Hmm the tests all pass but I ran it through the recognizer, the problem is that a blank space is detected above it. Spoke to @Tr4in about how the artefact filtering will be designed and it seems it will solve this issue 🎈  Blocked by #332

clemenshelm:
I don't think it's blocked. The problem is rather that the current `right_above` method will pick up a random word which is located somewhere above on the page, but not necessarily right above the word.

tamacodechi:
OH! It's picked up the first word to match the criteria which of course is right at the top of the page 🙈 Despotic limiting multiplier it is then 😂

tamacodechi:
Non despotic multiplier solution, heck yeah! 🙌 
I was going to do the quick fix and refactor it later but I think architecturally this makes more sense, especially since some other issues can reply on `in_the_same_column`

Tr4in (commented):
Perfect :sun_with_face:

lib/models/dimensionable.rb
```diff
@@ -46,5 +55,15 @@ def below(current)
...
     end
+
+    def in_same_column(word1, word2)
+      if word1.width > word2.width
+        word2_center = word2.left + (word2.width / 2)
+        word2_center >= word1.left && word2_center <= word1.right
+      else
+        word1_center = word1.left + (word1.width / 2)
+        word1_center >= word2.left && word1_center <= word2.right
+      end
```
clemenshelm: This looks quite complex. Wouldn't this condition suffice? `word1.left < word2.right && word2.left < word1.right` meaning that they are overlapping?

clemenshelm: This would pretty much replicate the logic of `on_same_line` 😄

clemenshelm: You could also use it in `right_below`.

tamacodechi: Wouldn't this also pick up things that are not in the same column? I was having a hard time wording it, so I just did a sketch 😅

But in the example the criteria are still met, but to me `above` is not in the same column at all.

It is a little more complex but I think the way it is now is quite robust and easy to refactor should our needs change. It's a little lengthy, but I also did this in exchange for legibility.

tamacodechi: We totally can! And I will apply it now 😉

clemenshelm: What about this case?

bildschirmfoto 2017-08-18 um 09 46 49

tamacodechi: I hadn't encountered an example where this was the case (whereby the definition they are in the same column but completely misaligned). 

Could you send me the bill ID? I'd love to look at the word attributes 🦑 

I feel like in this case it applies, but because it's in a table. I don't think we would want words (eg. in a paragraph) that are aligned like this to be in a column by our definition.

clemenshelm: It was a contrived example. Are all of these conditions you wrote necessary to fulfill the needs of the test data you used? To me this really looks like we're overcomplicating things. But if it's necessary for the data to work, I'm convinced 😄

clemenshelm:
Very nice! 👍

clemenshelm (approved):

PR #339, opened 2017-08-11
Increase timeout to allow lowe quality bills to be processed without …

### LINKED ISSUE #337: Bad quality bills make OCR very slow
Apparently it's a thing in Tesseract 4.0.0, fuzzy text makes everything slow! 🎉 

The best solution for now may be to raise the timeout and observe 👀 

Examples:
`dw63gm2okkaSAbESA.pdf`
`Q3Evi6BC7S5JBAJL9.pdf`
`N4JMu6aMQ5TafMp7k.pdf `

We can also use the benchmark module to see where things get slow.

tamacodechi:
…timing out

Closes #337 

🙈 Oops my commit message was too long!
Upped to 400 because just the recognition takes around 200+ and we need some buffer for the other processes as well

### DIFF
--- a/sidekiq.rb
+++ b/sidekiq.rb
@@ -29,7 +29,7 @@ class RecognitionWorker
       image_url: image_url,
       customer_vat_number: customer_vat_number
     )
-    timeout_in_secs = 200
+    timeout_in_secs = 400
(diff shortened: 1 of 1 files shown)


clemenshelm (approved):
Nice! 😼

PR #343, opened 2017-08-17
335 qr codes with different formats

### LINKED ISSUE #335: QR codes with data in different formats
Some QR codes contain underscores in the first couple of strings, which causes our data to be shifted over, and above all else **our dates are not calculated correctly and cause an error** 🙀 

> Some things to note: 
>1.  Both formats of QR code have 14 items
>2. The second string seems to always be "R1" by itself :wink:

Some bills in which this is the case:
- [x]  NjWHTCkMmkX925BvW.png
- [x]  oLEReFa2zTycKn5KW.jpg
- [x]  ga9GnkXgnGxesjQ8i.pdf

tamacodechi:
Closes #335 

Something I am unhappy with about this issue is that this shop (all examples came from the same shop), seems to have gotten the format of the data within the qr code wrong because they have placed the 13% vat rate second instead of 10% 🐝


lib/qr_decoder.rb
```diff
@@ -24,8 +17,17 @@ def decode_qr_code
...
-    prices_and_vats = VAT_RATES.zip(prices).to_h
+    if all_data[1] == 'R1'
+      vat_rates = [20, 13, 10, 0, nil]
+      prices = all_data[6..10].map { |p| BigDecimal.new(p.sub(',', '.')) }
+      date_in_qr = DateTime.strptime(all_data[5], '%Y-%m-%d').strftime('%Y-%m-%d')
+    else
+      vat_rates = [20, 10, 13, 0, nil]
+      prices = all_data[5..9].map { |p| BigDecimal.new(p.sub(',', '.')) }
+      date_in_qr = DateTime.strptime(all_data[4], '%Y-%m-%d').strftime('%Y-%m-%d')
+    end
```
clemenshelm: This would make a nice little method of its own:

```ruby
def prices_and_date_parts(qr_code_parts)
  if qr_code_parts[1] == 'R1'
    { prices: qr_code_parts[6..10], date: qr_code_parts[5] }
  else
    { prices: qr_code_parts[5..9], date: qr_code_parts[4] }
  end
end
```

It would resolve all the duplication here.

clemenshelm (changes requested):
Please DRY 🙂

lib/qr_decoder.rb
tamacodechi: Looks good, I'll do something like this then ✌️ 

I'm always a little unsure of compartmentalising, but would you recommend erring on the side of doing it and then going back and fixing it if it's unnecessary?

lib/qr_decoder.rb
```diff
@@ -57,4 +53,20 @@ def qr_codes
...
+        prices: qr_code_parts[6..10].map { |p| BigDecimal.new(p.sub(',', '.')) },
+        date: qr_code_parts[5]
+      }
+    else
+      {
+        vat_rates: [20, 10, 13, 0, nil],
+        prices: qr_code_parts[5..9].map { |p| BigDecimal.new(p.sub(',', '.')) },
+        date: qr_code_parts[4]
+      }
+    end
```
tamacodechi: I am not 100% sure making a method of it's own was that great a success... 
- Accessing the data makes the lines considerably longer. 
- Unless I am doing something wrong, the prices have to processed at the same time as they are extracted or they are converted into strings, so this has to be repeated (`.map { |p| BigDecimal.ne....`)

Perhaps moving the date processing outside the conditional flow would be a better step forward? 

Of course I don't mind the method otherwise 🎉 I think it makes the `decode_qr_code` cleaner anyway

clemenshelm: The `.map { |p| BigDecimal.new(p.sub(',', '.')) }` could also be outside of the method.

Regarding length: You can destructure the resulting hash by writing
`date, prices, vat_rates = filtered_data.values_at(:date, :prices, :vat_rates)`

lib/qr_decoder.rb
```diff
@@ -58,15 +48,31 @@ def determine_data_format_and_set_data(qr_code_parts)
...
+
+    prices_and_vats = vat_rates.zip(formatted_prices).to_h
+    prices_present = prices_and_vats.select { |_vat, price| price.positive? }
+
+    prices_present.map do |vat, price|
+      {
+        total: price,
+        vatRate: vat
+      }
+    end
```
clemenshelm, line 76: Very nice. You can also chain the method calls like

```ruby
vat_rates
  .zip(…)
  .select(…)
  .map(…)
```

This way you don't have to come up with variable names all the time. The only condition is that readibility shouldn't suffer by it. In this case I think it would even improve readability because it would emphasize the "data processing pipeline".

You don't need to change it anymore, though, this is just a suggestion for the future.

clemenshelm (approved):
Good job! 🥇

lib/qr_decoder.rb
tamacodechi: Oh that does look considerably better... 🙈 I will keep this in mind and fix it when something else comes around on the QR code decoder

PR #347, opened 2017-08-18
332 wrong text box dimensions

### LINKED ISSUE #332: Wrong text-box dimensions

Tr4in:
Closes #332


lib/image_processor.rb
```diff
@@ -61,13 +60,22 @@ def apply_background(color)
...
   end
 
   def normalize
     process_image(&:normalize)
   end
 
+  def update_width_and_height
+    @image_width = @image.columns
+    @image_height = @image.rows
+  end
```
tamacodechi, line 77: 💖

lib/models/word.rb
```diff
@@ -18,4 +18,9 @@ def follows(previous_word)
     # space: #{max_space_width} => last: #{previous_word.left}"
     previous_word.right + max_space_width >= left
   end
+
+  def self.filter_out_artifacts
+    # Filter Conditions
```
tamacodechi: I don't think this comment is necessary, if the code is good, it explains itself :)

lib/models/word.rb
```diff
@@ -18,4 +18,9 @@ def follows(previous_word)
...
+    where(text: [' ', '*'])&.destroy
```
tamacodechi: Here you're removing the artefact words completely from our saved words?
🤔  @clemenshelm  do you think it's okay like this or should we be creating a subset of "desired" words instead?

lib/detectors/vat_number_detector.rb
```diff
@@ -33,7 +33,7 @@ def find_vat_numbers(regex, after_each_word: nil)
         terms = Word.where(
           'bottom - top <= ?', median_height * scale_factor
         ).map do |word|
-          if term_stale || (last_word && !word.follows(last_word))
+          if term_stale
```
tamacodechi: Have you tried playing around with the `.follows` method instead of completely removing this here? I'm sure that condition was implemented for a reason 😉

lib/bill_recognizer.rb
```diff
@@ -98,8 +98,12 @@ def recognize
     end
 
     recognize_words(png_file)
-    filter_words
+    Word.filter_out_artifacts
+
+    # ****** PRINT ALL WORDS ******
+    # print_words
```
tamacodechi: I think the method name "print_words" or `print_all words` or `print_all_detected words` is clear enough, the comment above is not necessary 😊

lib/bill_recognizer.rb
```diff
@@ -136,16 +140,18 @@ def recognize_words(png_file)
...
+  def print_words
+    puts Word.map { |word|
+      "
+      text: \'#{word.text}\',
+      left: #{word.left},
+      right: #{word.right},
+      top: #{word.top},
+      bottom: #{word.bottom}
+      "
+    }
```
tamacodechi, line 153: 👍

spec/models/word_spec.rb
```diff
@@ -87,4 +87,34 @@
 
     expect(word2.follows(word1)).to eq true
   end
+
+  it 'filters out words with blank space' do
```
tamacodechi: You also filtered out an asterisk here, you should probably rename this to also describe that.

tamacodechi (changes requested):
Awesome work 🌟 I think a review from @Thomas1e would be very useful here! Have you considered adding a spec that shows why the deskew cropping and updating of the dimensions is necessary? When we backtrack and look at the code it might not be so clear why we do that 🤔 A spec would be awesome 😋

lib/models/word.rb
clemenshelm: It's ok. If we need these words at a later point, we can still save them. Personally I'd have filtered them out even before inserting them into the database, but I think it doesn't make much of a difference.

lib/models/word.rb
```diff
@@ -13,14 +13,13 @@ def next
   end
 
   def follows(previous_word)
-    max_space_width = previous_word.height * 1.79
+    max_space_width = previous_word.height * 1.80
```
tamacodechi, line 16: 0.01 😂

lib/bill_recognizer.rb
```diff
@@ -142,7 +141,7 @@ def recognize_words(png_file)
     create_words_from_hocr(hocr_doc)
   end
 
-  def print_words
+  def print_all_detected_words
```
tamacodechi: Just saw that it's after the words are filtered out, maybe just `print_all_words` since detected words includes the artefacts

spec/detectors/vat_number_detector_spec.rb
```diff
@@ -828,7 +828,7 @@
     expect(vat_numbers.map(&:to_s)).to eq ['ATU65315367']
   end
 
-  it 'detects vats' do
+  it 'detects vats with high space width' do
```
tamacodechi: Can you label this spec please? 😊

tamacodechi (changes requested):

tamacodechi (approved):

PR #351, opened 2017-08-22
Notify Rollbar about deployment for auto resolving and increment version number

### LINKED ISSUE #350: Notify rollbar about deployment

tamacodechi:
Closes #350 I just realized I can't actually run it through until I put this into master and commit it thanks to line 63 🙈

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -62,7 +62,7 @@ task :git_check do
   branch = `branch_name=$(git symbolic-ref HEAD 2>/dev/null); branch_name=${branch_name##refs/heads/}; echo ${branch_name:-HEAD}`.strip
   abort "You have checked out the #{branch} branch, please only deploy from the master branch!" unless branch == 'master'
 
-  abort "You have unstaged or uncommitted changes! Please only deploy from a clean working directory!" if `git status --porcelain`.present?
+  abort "You have unstaged or uncommitted changes! Please only deploy from a clean working directory!" unless `git status --porcelain`.empty?
 end
 
 desc 'Increment recognizer version number'
--- a/Rakefile
+++ b/Rakefile
@@ -74,7 +74,7 @@ task :increment_version => [:git_check] do
 
   sh "git add .
 
-      git commit -m "Increase version number"
+      git commit -m 'Increase version number'
 
       git push origin master"
 end
--- a/Rakefile
+++ b/Rakefile
@@ -108,7 +108,7 @@ task :restart_task do
 end
 
 desc 'Notify Rollbar about deployment so it can autoresolve all errors'
-task :notify_rollbar => [:push_image, :restart_task] do
+task :notify_rollbar do
   require 'YAML'
   data = YAML.load_file "lib/version.yml"
   recognizer_version = data["Version"]
@@ -117,7 +117,7 @@ task :notify_rollbar => [:push_image, :restart_task] do
 end
 
 desc 'Increments recognizer version number and deploys newest version'
-task :deploy => [:notify_rollbar] do
+task :deploy => [:push_image, :restart_task, :notify_rollbar] do
   p "Newest recognizer version successfully deployed!✌️"
 end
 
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -89,11 +103,22 @@ task :restart_task do
   latest_revision = all_revision_numbers.map {|num| num.to_i}.sort.last
 
   sh "aws ecs run-task --cluster ChillBill --task-definition ecscompose-recognizer:#{latest_revision} --count 1 --region eu-central-1"
+
+  p "The recognizer task has successfully been restarted!🦑"
+end
+
+desc 'Notify Rollbar about deployment so it can autoresolve all errors'
+task :notify_rollbar => [:push_image, :restart_task] do
```
clemenshelm: Why does notifying Rollbar require deployment?

Rakefile
```diff
@@ -89,11 +103,22 @@ task :restart_task do
...
+  data = YAML.load_file "lib/version.yml"
+  recognizer_version = data["Version"]
+  access_token = ENV['ROLLBAR_ACCESS_TOKEN']
+  sh "curl https://api.rollbar.com/api/1/deploy/ -F access_token=#{access_token} -F environment=production -F revision=#{recognizer_version}"
 end
 
 desc 'Increments recognizer version number and deploys newest version'
-task :deploy => [:push_image, :restart_task] do
-  p "Newest recognizer version successfully deployed!"
+task :deploy => [:notify_rollbar] do
```
clemenshelm: It should rather be like `task :deploy => [:push_image, :restart_task, :notify_rollbar]`

clemenshelm (changes requested):

Rakefile
tamacodechi: Because we don't want to auto resolve all errors unless we actually redeploy, do we?

Rakefile
tamacodechi: Either way works, but the reason I've written it like this is that it's a little more clear that rollbar won't be notified unless deployment takes place as the steps of redeployment are required by the `:notify_rollbar` task explicitly. I'm not fussed, should I change it? 😊

Rakefile
clemenshelm: But this implicates that when you only want to notify Rollbar, another deployment is triggered.

Rakefile
clemenshelm: If it's written as the last step this is also clear because if a previous step fails, the next steps won't be executed. Please change it. 😉

Rakefile
tamacodechi: Absolutely true! I just don't think it's logical that we notify rollbar unless we deploy, isn't that something we never want to allow?

Rakefile
tamacodechi: Okay will do 👍 Yep, it's just implicit when they're all required by `:deploy`

clemenshelm: By the logic you explained, `:restart_task` should also depend on `:push_image` because we don't want to restart a task unless the image was pushed. But I think it's not very clear to define steps like this 😉

tamacodechi: Yep, but the difference there is that we often do want to restart the task without pushing the image, like when the recognizer crashes, whereas we never ever want to notify rollbar of a "deployment" unless we actually do redeploy.

But I think I may have been over-designing 😅

clemenshelm: I see your point, but if we ever need to do it for whatever reason (although we "never ever" should, but Rollbar was down just when we deployed and actually that gave us a reason) we would be quite suprised that the whole app was redeployed even though we just wanted to inform Rollbar.

tamacodechi: 😂 I've already changed it, but for the sake of that edge case, shouldn't we then do a cleanup via Rollbar and not the recognizer? We'd anyway be on Rollbar to see when it's back up 😅

clemenshelm: Yeah probably. But this way we CAN do it also on the recognizer 😉

clemenshelm (approved):

tamacodechi:
Already noticed something much worse, the increase version push should be part of the deploy task, even if it's not intuitive, because if deploy fails, it still pushes a version number change...

Issue #342, opened 2017-08-17
Kto. recognized as price


Thomas1e:
In 9SALLXySPgesCxkCs.pdf the Kto. "13 300" is recognized as price.

tamacodechi:
Closed by #353

Issue #341, opened 2017-08-17
BLZ recognized as price


Thomas1e:
In 9SALLXySPgesCxkCs.pdf the BLZ "32 250" is recognized as price.

tamacodechi:
Closed by #353

Issue #237, opened 2017-04-29
VAT number discovered as price


clemenshelm:
In bill 2Q3jZFWCscBgqov2q.pdf a VAT number is discovered as price:

"77 967" is recognized as 77967.

<img width="333" alt="screen shot 2017-04-29 at 21 20 07" src="https://cloud.githubusercontent.com/assets/141632/25558212/27aa53c0-2d22-11e7-87f2-5679913d8e56.png">

tamacodechi:
Closed by #353

Issue #230, opened 2017-04-26
UST Number recognized as price


clemenshelm:
In bill 2JjDSP33hX6XMngHd.pdf, the UST number is recognized as price:

<img width="261" alt="screen shot 2017-04-26 at 20 06 52" src="https://cloud.githubusercontent.com/assets/141632/25449447/11122a10-2abc-11e7-85a3-925d182608cd.png">

"68 603" is recognized as price 68 603.

tamacodechi:
Closed by #353

PR #353, opened 2017-08-23
304 phone number recognized as price

### LINKED ISSUE #302: adress recognized as prize
In JBBfpB9Fypf54hTz2.pdf and HaPoDhJqjpWRavh3h.pdf the adress 14 302 is recognized as price 14302,00.

jbbfpb9fypf54htz2

Other examples:
- iTGXBQoDzEsj8H54c

### LINKED ISSUE #304: phone number recognized as prize
In gNPBm9p7ttJJFgCdY.pdf part of the phone number is recognized as prize.
gnpbm9p7ttjjfgcdy

Other examples:
- y5PuTZzuWvYJ7bQaS.pdf
- Xrx6yha9FMsBdNGAm.pdf (92 409)

tamacodechi:
Finally a way to limit that hungarian regex 🤣 

Closes #304 

I do think that the word having a new attribute could solve many problems in the future but this fix works for this case now.

Edit:
Now that it's been merged and we can test from master:
Closes #302

### DIFF
--- a/lib/detectors/currency_detector.rb
+++ b/lib/detectors/currency_detector.rb
@@ -9,7 +9,7 @@ class CurrencyDetector
   CNY_SYMBOLS = %w(CNY).freeze
   SEK_SYMBOLS = %w(SEK).freeze
   GBP_SYMBOLS = %w(GBP £).freeze
-  HUF_SYMBOLS = %w(HUF Ft.).freeze
+  HUF_SYMBOLS = %w(HUF Ft. Ft).freeze
   HRK_SYMBOLS = %w(HRK).freeze
   ALL_SYMBOLS = EUR_SYMBOLS + USD_SYMBOLS + HKD_SYMBOLS +
                 CHF_SYMBOLS + CNY_SYMBOLS + SEK_SYMBOLS + GBP_SYMBOLS +
--- a/lib/detectors/price_detector.rb
+++ b/lib/detectors/price_detector.rb
@@ -2,6 +2,7 @@
 require 'ostruct'
 require 'bigdecimal'
 require_relative '../models/price_term'
+require_relative '../detectors/currency_detector'
 
 class PriceDetector
   PRICE_REGEX = /(-?[1-9]{1}\d{0,3}|0)([\.,]\d{3})?[,\.](\d{2}|-)/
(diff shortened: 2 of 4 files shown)


tamacodechi:
Nicely spotted @Tr4in ✌️ 
Probably fixes #302 and now that I found it #224 (which seems to be a duplicate of 302 with more examples) as well 🎉 

#230, #237, #341 and #342 would be fixed too. I'm pretty sure they will because I've tested them.

lib/detectors/currency_detector.rb
```diff
@@ -9,7 +9,7 @@ class CurrencyDetector
   CNY_SYMBOLS = %w(CNY).freeze
   SEK_SYMBOLS = %w(SEK).freeze
   GBP_SYMBOLS = %w(GBP £).freeze
-  HUF_SYMBOLS = %w(HUF Ft.).freeze
+  HUF_SYMBOLS = %w(HUF Ft. Ft).freeze
```
Tr4in, line 12: I think Ft.? would work too

tamacodechi: Hmm I rewrote one spec and wrote the other and it wouldn't pick it up unless it was like this, so I'm not sure what you mean 😅 

EDIT: Now that I understand, these are single quoted strings not written as regexes so unless I change it into `%r{...}`using `Ft.?` wouldn't work

Since we're using single quoted strings everywhere else, even in the limiting of the hungarian price regex, I think it would be cleaner to keep it like this. What do you think? I think if we have to add one more then then maybe it may be worth it 😉

Tr4in: Keep it :new_moon_with_face:

lib/detectors/price_detector.rb
```diff
@@ -26,11 +26,13 @@ def self.filter
 
     end_word_with_space = ->(term) { term.text += ' ' }
 
-    find_prices(
-      HUNGARIAN_PRICE_REGEX,
-      after_each_word: end_word_with_space,
-      max_words: 2
-    )
+    unless Word.where(text: ['Ft.', 'Ft', 'HUF']).empty?
```
clemenshelm: Please use `CurrencyDetector::HUF_SYMBOLS` instead of duplicating the array.

clemenshelm (changes requested):

lib/detectors/price_detector.rb
tamacodechi: Didn't think of that, thank you! 👏

clemenshelm (approved):

clemenshelm:
Meeeerged! @Althaire can you close all other issues as well which are closed for sure?

tamacodechi:
Oh hahaha I guess you were writing that as I was walking over to ask. Wiiiiillll doooooo ✌️ right now 🍡

PR #356, opened 2017-08-23
changed below method

### LINKED ISSUE #212: Below and right_below are similar or redundant
Right now we have below which detects things very much below a word. We NEED right_below for the invoice number recognition. Identify the discrepancies so the code doesn't have repetition and still suits our needs.

Tr4in:
Closes #212 

If you have questions ask @vladofilipic :sun_with_face:

### DIFF
--- a/lib/models/dimensionable.rb
+++ b/lib/models/dimensionable.rb
@@ -48,10 +48,12 @@ module Dimensionable
       everything_above.last
     end
 
-    def below(current)
-      all.find do |lower|
-        lower.right > current.left && lower != current
+    def last_below(current)
+      everything_above = all.select do |lower|
+        lower.top > current.top && lower.bottom > current.bottom && current != lower &&
+        current.left < lower.left
       end
+      everything_above.last
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/models/dimensionable.rb
```diff
@@ -48,10 +48,12 @@ def right_above(current)
...
 
-    def below(current)
-      all.find do |lower|
-        lower.right > current.left && lower != current
+    def last_below(current)
+      everything_above = all.select do |lower|
+        lower.top > current.top && lower.bottom > current.bottom && current != lower &&
+          current.left < lower.left
       end
+      everything_above.last
```
tamacodechi: I'm not sure the variable should be called `everything_above` 😂 

I think a simpler solution that does exactly what you're describing here would be something like:
```ruby
def bottom_most(current)
  everything_below = all.select do |lower|
    in_same_column(current, lower)
  end
  everything_below.last
end
```
Something like this 😉

tamacodechi (changes requested):
🎉

tamacodechi (approved):
👏 Awesome 👏

lib/models/dimensionable.rb
```diff
@@ -48,10 +48,11 @@ def right_above(current)
       everything_above.last
     end
 
-    def below(current)
-      all.find do |lower|
-        lower.right > current.left && lower != current
+    def bottom_most(current)
```
tamacodechi, line 51: Wonderful, such a clear name ✌️

Issue #284, opened 2017-06-26
VAT percent recognized as price


clemenshelm:
In bill 7J4sSvSesAHz5bHyT.pdf the VAT percentage is recognized as price.

![Bildschirmfoto 2017-06-26 um 11.16.28.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/57e2169a778e3d0f005c2d9c/125516c66e82c728ace21e0d46ca9d96279180a186fbb245ab49ffd47b1676246454f56622d3a33db2022a50415e48b243504d58eeb9cb74fce36f34985619e8d6385d75c06914f9b1874de900eb278b4fd6e4be8b19c8fa2137a5dad11b61435c52dcbd4f3fb29a49c259f242e94953ac15326be274306e0a.png)

Thomas1e:
Another example: hWtW3cWFwy842Ribr.pdf (I am not sure what "Auftragsvorlage" in this case means, but it increases the netto price). It was also recognized as price.

Another example: BStaGs5GAwHCaYLaz.pdf

![bildschirmfoto vom 2017-08-17 12-55-27](https://user-images.githubusercontent.com/26839274/29409583-354a1b10-834d-11e7-8ee1-c2b76b2f714c.png)

Issue #241, opened 2017-04-30
Hours recognized as prices


clemenshelm:
In bill 2tMo5XaJRTcPwG55r.pdf, an amount of hours is recognized as price:

<img width="154" alt="screen shot 2017-04-30 at 10 58 08" src="https://cloud.githubusercontent.com/assets/141632/25563070/141f12d8-2d94-11e7-96fe-28a697af9c89.png">

Thomas1e:
Another example BStaGs5GAwHCaYLaz.pdf

![bildschirmfoto vom 2017-08-17 12-55-27](https://user-images.githubusercontent.com/26839274/29409364-238c144c-834c-11e7-9707-25138eac0b04.png)

Issue #240, opened 2017-04-30
Discount percent recognized as price


clemenshelm:
In bill 2qZF2tx7qPXzTPFLt.pdf the discount percentage is recognized as prices:

<img width="96" alt="screen shot 2017-04-30 at 10 42 35" src="https://cloud.githubusercontent.com/assets/141632/25562925/c2a3aee8-2d91-11e7-99dc-8488ee831059.png">

Thomas1e:
Another example BStaGs5GAwHCaYLaz.pdf

![bildschirmfoto vom 2017-08-17 12-55-27](https://user-images.githubusercontent.com/26839274/29409505-c37729b0-834c-11e7-9a57-953ef592c54b.png)

Issue #229, opened 2017-04-26
Bank account number recognized as price


clemenshelm:
In bill 2HtXGWoynP3sFMvFH.pdf, part of the bank account number is recognized as price:

<img width="136" alt="screen shot 2017-04-26 at 20 01 40" src="https://cloud.githubusercontent.com/assets/141632/25449203/55de31b2-2abb-11e7-8398-b838e1e93803.png">

"27 900" is recognized as price 27900.

Also in bill 2X3ybPNqDbuwH5dSX.pdf the bank account number "2.654.580" is recognized as price 265458:

<img width="141" alt="screen shot 2017-04-29 at 21 48 05" src="https://cloud.githubusercontent.com/assets/141632/25558407/52968190-2d26-11e7-8840-d4eb64c35a8c.png">

Tr4in:
On bill 2X3ybPNqDbuwH5dSX.pdf: 5,66 got recognized as price but its percent

PR #357, opened 2017-08-23
first commit

### LINKED ISSUE #299: Different units recognized as prices
In bill KGEFGTvogD5Ly9jTT.pdf many units are recognized as prices.

bildschirmfoto 2017-07-04 um 19 51 50

Tr4in:
Closes #299 

This will punch #229, #240, #241 and #284 too!

### DIFF
--- a/lib/detectors/price_detector.rb
+++ b/lib/detectors/price_detector.rb
@@ -7,7 +7,7 @@ require_relative '../detectors/currency_detector'
 class PriceDetector
   PRICE_REGEX = /(-?[1-9]{1}\d{0,3}|0)([\.,]\d{3})?[,\.](\d{2}|-)/
   PREFIX_CURRENCY_REGEX = /(€|EUR)/
-  ALLOWED_PREFIX_REGEX = /(?:^|[^\d,A-Za-z\.])/
+  ALLOWED_PREFIX_REGEX = /(?:^|[^\d,A-Za-z\.-])/
   DECIMAL_PRICE_REGEX =
     /#{ALLOWED_PREFIX_REGEX}(#{PREFIX_CURRENCY_REGEX}?#{PRICE_REGEX})$/
   WRITTEN_PRICE_REGEX = /(\d+ Euro)/
--- a/lib/models/price_term.rb
+++ b/lib/models/price_term.rb
@@ -9,7 +9,13 @@ class PriceTerm < Sequel::Model
   include Dimensionable
 
   def valid?
-    super && Word.right_after(self)&.text != 'kg' && Word.right_above(self)&.text != 'kg'
+    super && !pass_unit_check
+    # binding.pry
+  end
+
+  def pass_unit_check
+    Word::UNITS.include?(Word.right_after(self)&.text) ||
+      Word::UNITS.include?(Word.right_above(self)&.text)
   end
 
   def to_d
--- a/lib/models/price_term.rb
+++ b/lib/models/price_term.rb
@@ -9,12 +9,12 @@ class PriceTerm < Sequel::Model
   include Dimensionable
 
   def valid?
-    super && !pass_unit_check
+    super && pass_unit_check
   end
 
   def pass_unit_check
-    Word::UNITS.include?(Word.right_after(self)&.text) ||
-      Word::UNITS.include?(Word.right_above(self)&.text)
+    !Word::UNITS.include?(Word.right_after(self)&.text) &&
+      !Word::UNITS.include?(Word.right_above(self)&.text)
   end
 
   def to_d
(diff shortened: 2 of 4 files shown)


lib/models/price_term.rb
```diff
@@ -9,7 +9,13 @@ class PriceTerm < Sequel::Model
   include Dimensionable
 
   def valid?
-    super && Word.right_after(self)&.text != 'kg' && Word.right_above(self)&.text != 'kg'
+    super && !pass_unit_check
+    # binding.pry
```
tamacodechi: 👀

lib/models/word.rb
```diff
@@ -8,6 +8,8 @@ class Word < Sequel::Model
   include Dimensionable
   include Logging
 
+  UNITS = %w(kg % Stunden SK PA SG ST TO KG M3).freeze
```
tamacodechi, line 11: Yay 💖 I love that we're doing it this way :)

spec/detectors/price_detector_spec.rb
```diff
@@ -793,6 +793,195 @@
...
+      left: 0.7889397905759162,
+      right: 0.8089005235602095,
+      top: 0.6089669516986365,
+      bottom: 0.6170556967876126
+    )
+
+    prices = PriceDetector.filter
+    expect(prices.map(&:text)).to be_empty
+  end
+
```
tamacodechi, line 1002: 👍  
But remember when you label a bill you can also just write:
 `# From <bill-id>.<extension>"`

lib/detectors/price_detector.rb
```diff
@@ -7,7 +7,7 @@
 class PriceDetector
   PRICE_REGEX = /(-?[1-9]{1}\d{0,3}|0)([\.,]\d{3})?[,\.](\d{2}|-)/
   PREFIX_CURRENCY_REGEX = /(€|EUR)/
-  ALLOWED_PREFIX_REGEX = /(?:^|[^\d,A-Za-z\.])/
+  ALLOWED_PREFIX_REGEX = /(?:^|[^\d,A-Za-z\.-])/
```
tamacodechi: Which spec needs this? 🤔

tamacodechi (changes requested):
Awesome! I think only one tiny thing, but it's perfect 👍

lib/detectors/price_detector.rb
Tr4in: I put it there because the spec **does not detect prices in a long weird word** which was not merged AND there was a bill with the same problem

lib/models/price_term.rb
```diff
@@ -9,7 +9,12 @@ class PriceTerm < Sequel::Model
   include Dimensionable
 
   def valid?
-    super && Word.right_after(self)&.text != 'kg' && Word.right_above(self)&.text != 'kg'
+    super && !pass_unit_check
+  end
+
+  def pass_unit_check
```
tamacodechi: This should probably be called `unit_check` because above we're saying `!pass_unit_check` which literally means "not pass unit check"... which it actually _is_ passing! 😁 Or change `pass_unit_check` to actually make sure that they do not include the unit words.

I actually like the name `pass_unit_check` so I would prefer the second fix, but it's up to how you prefer 😉

spec/detectors/price_detector_spec.rb
```diff
@@ -811,6 +811,195 @@
...
+      right: 0.44404450261780104,
+      top: 0.5289084181313598,
+      bottom: 0.5374653098982424
+    )
+
+    prices = PriceDetector.filter
+    expect(prices.map(&:text)).to be_empty
+  end
+
+  it 'does not detect prices' do
```
tamacodechi: I hope it detects price! 🙈  I think you mean it doesn't detect hours as prices, right? :)

spec/detectors/price_detector_spec.rb
```diff
@@ -811,6 +811,195 @@
...
+      right: 0.3829300196206671,
+      top: 0.5102890173410405,
+      bottom: 0.5193063583815029
+    )
+
+    prices = PriceDetector.filter
+    expect(prices.map(&:text)).to be_empty
+  end
+
+  it 'does not detect multi units' do
```
tamacodechi: "multiple units as prices"
🙈 🙈 So picky 😂 😂

tamacodechi (changes requested):
Can't wait to merge this 🎉 Just a couple of tiny things 😅

tamacodechi (approved):
Ideally the spec where you are testing for multiple units would be broken into individual specs so they can be more descriptive. From that one spec name we can't really know which of the units are being tested against 😉 

👏 👏  Good job! 👍

PR #360, opened 2017-08-29
333 missing prices

### LINKED ISSUE #333: missing prices
- In the bill 2BWLcAHqKAJTzwSDC.jpg a lot of prices are not in detected and therefore are missing in the yml file.

Prices that are detected: 5.34;  0.49;   4.85;  300

bildschirmfoto vom 2017-08-08 16-22-26

- In fRN7LRFs8ZRyyeFdE.pdf the MwSt value was not detected, but it should be.

- In FvK6uW7F9A62f4YGZ.pdf the value "41.02" should be detected 4 times, but was only detected twice.

- In C5sri9hxpbDhha68D.png the value "23,55" should be detected 2 times, but was not detected

- In rDxLnivxoXQw9nWa7.pdf the values "431.25" and "86.25" are not detecetd

- In LcW5EsYpEX7FTsUF3.jpg the value "8.29" is missing

### LINKED ISSUE #361: wrong text-box dimensions
We still have some "artifacts" recognized as text. To get the best possible results, we need to filter them out.

Examples:
- cN6h5XBq6howTHQwa.pdf (text_box_top = 0, text_box_left = 0,text_box_right and text_box_bottom are to big)

Tr4in:
Closes #333 
Closes #361 

I think that the syntax has to be corrected!

### DIFF
--- a/lib/bill_recognizer.rb
+++ b/lib/bill_recognizer.rb
@@ -100,7 +100,7 @@ class BillRecognizer
     recognize_words(png_file)
     Word.filter_out_artifacts
 
-     print_all_words
+    # print_all_words
 
     filter_words
     calculate_text_box
--- a/lib/bill_recognizer.rb
+++ b/lib/bill_recognizer.rb
@@ -100,7 +100,7 @@ class BillRecognizer
     recognize_words(png_file)
     Word.filter_out_artifacts
 
-    # print_all_words
+     print_all_words
 
     filter_words
     calculate_text_box
--- a/lib/bill_recognizer.rb
+++ b/lib/bill_recognizer.rb
@@ -98,7 +98,6 @@ class BillRecognizer
     end
 
     recognize_words(png_file)
-    Word.filter_out_artifacts
 
     # print_all_words
 
@@ -139,6 +138,7 @@ class BillRecognizer
 
     hocr_doc = Nokogiri::HTML(hocr)
     create_words_from_hocr(hocr_doc)
+    Word.filter_out_artifacts
   end
 
   def print_all_words
--- a/lib/detectors/date_detector.rb
+++ b/lib/detectors/date_detector.rb
@@ -33,12 +33,11 @@ class DateDetector
     find_dates(reduced_words, LONG_HUNGARIAN_DATE_REGEX, max_words: 1)
 
     DateTerm.order(:first_word_id)
-    binding.pry
   end
 
   def self.filter_out_interefering_date_terms
     words = find_dates(Word.all, LONG_SLASH_DATE_REGEX, max_words: 1)
-    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 3)
+    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 6)
     words += find_dates(Word.all, LONG_YEAR_SLASH_REGEX, max_words: 1)
     Word.all - words
   end
--- a/lib/detectors/date_detector.rb
+++ b/lib/detectors/date_detector.rb
@@ -33,11 +33,12 @@ class DateDetector
     find_dates(reduced_words, LONG_HUNGARIAN_DATE_REGEX, max_words: 1)
 
     DateTerm.order(:first_word_id)
+    binding.pry
   end
 
   def self.filter_out_interefering_date_terms
     words = find_dates(Word.all, LONG_SLASH_DATE_REGEX, max_words: 1)
-    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 2)
+    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 3)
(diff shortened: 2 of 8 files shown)


lib/detectors/date_detector.rb
```diff
@@ -37,7 +37,7 @@ def self.filter
 
   def self.filter_out_interefering_date_terms
     words = find_dates(Word.all, LONG_SLASH_DATE_REGEX, max_words: 1)
-    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 2)
+    words += find_dates(Word.all, SHORT_PERIOD_DATE_REGEX, max_words: 6)
```
tamacodechi, line 40: 👍

lib/detectors/price_detector.rb
```diff
@@ -17,13 +19,28 @@ class PriceDetector
   def self.filter_out_quantity_column
     %w(Menge Anz.).each do |quantity_text|
       quantity = Word.first(text: quantity_text)
-      PriceTerm.where { right <= quantity.right }.destroy if quantity
+      next unless quantity
+      PriceTerm.map do |term|
```
tamacodechi: Is `.map` necessary? I think `.each` should be enough, or?

spec/detectors/date_detector_spec.rb
```diff
@@ -775,8 +775,60 @@
     expect(date_strings(dates)).to be_empty
   end
 
+  it 'detects dates with max_words 6' do
```
tamacodechi: "consisting of 6 words"

It's a little odd to have the attribute name there 😅

lib/models/word.rb
```diff
@@ -8,14 +8,14 @@ class Word < Sequel::Model
...
-  UNITS = %w(kg % Stunden SK PA SG ST TO KG M3).freeze
+  UNITS = %w(kg % Stunden SK PA SG ST TO KG M3 MWST).freeze
 
   def next
     Word[id + 1]
   end
 
   def follows(previous_word)
-    max_space_width = previous_word.height * 1.80
+    max_space_width = previous_word.height * 10.0
```
tamacodechi, line 18: Can't wait 'til we stop using `.follows` 😅 This fixes the current problem though. Have you noticed any change in acceptance test results? Like the terms that get picked up and so on

lib/detectors/price_detector.rb
```diff
@@ -17,13 +19,28 @@ class PriceDetector
...
+        term.destroy if in_same_column(quantity, term)
```
tamacodechi: 👏 👏 👏

lib/detectors/price_detector.rb
```diff
@@ -17,13 +19,28 @@ class PriceDetector
...
+  end
+
+  def self.filter_out_dates
+    PriceTerm.map do |term|
+      first_word = Word.right_after(term)
+
+      next unless first_word && first_word.text.match(/\./)
+      second_word = Word.right_after(first_word)
+      term.destroy if second_word &&
+                      second_word.text.match(/^(\d{2})$/)
```
tamacodechi: It might be nice to extend dimensionable with a method with this behaviour 😊 I can see many cases where we could use this 😋 It's also loosely defined here and it feels like it's not really behaviour that a detector should define, right?

spec/detectors/price_detector_spec.rb
```diff
@@ -1018,6 +1018,159 @@
     expect(prices.map(&:text)).to eq ['2,17€']
   end
 
+  it 'detects prices which are not in between of quantity' do
```
tamacodechi: "which are not in part of a quantity"

tamacodechi (changes requested):
Very nice 😉

lib/models/dimensionable.rb
```diff
@@ -48,6 +48,17 @@ def right_above(current)
       everything_above.last
     end
 
+    def match_two_words_after(current, regex1, regex2)
```
tamacodechi: `.two_word_after_matches?`

tamacodechi: This returns true or false 🎉

lib/bill_recognizer.rb
```diff
@@ -139,6 +138,7 @@ def recognize_words(png_file)
 
     hocr_doc = Nokogiri::HTML(hocr)
     create_words_from_hocr(hocr_doc)
+    Word.filter_out_artifacts
   end
```
tamacodechi, line 142: "Closes #361 " 🥇

spec/detectors/date_detector_spec.rb
```diff
@@ -775,8 +775,60 @@
     expect(date_strings(dates)).to be_empty
   end
 
+  it 'detects dates which consists of 6 words' do
+    # From C5sri9hxpbDhha68D.png
+
```
tamacodechi: 👀

tamacodechi (changes requested):
🌮

tamacodechi (approved):
👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #366, opened 2017-09-01
Raise timeout

### LINKED ISSUE #359: Evaluate bill and possibly raise timeout
The bill `ysPgNkdA9bTuYtGRR.pdf` has caused 450 timeout errors in the last 7 days. It's the only bill left still causing timeout errors. Use the benchmark module to evaluate how much time it needs and see whether it's worth it to raise the timeout.

tamacodechi:
benchmarked it and it took about 460 seconds 🐢 

Closes #359 

UPDATE: 
> Will keep an eye on rollbar, but it seems that the bill has gotten reprocessed. Maybe there are so many new bills that it hasn't been able to get requeued again 😂

### DIFF
--- a/sidekiq.rb
+++ b/sidekiq.rb
@@ -29,7 +29,7 @@ class RecognitionWorker
       image_url: image_url,
       customer_vat_number: customer_vat_number
     )
-    timeout_in_secs = 400
+    timeout_in_secs = 500
     bill_attributes = Timeout.timeout(timeout_in_secs) { recognizer.recognize }
     bill_attributes[:id] = id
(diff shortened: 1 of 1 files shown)


clemenshelm:
Did you find out what's taking so long?

tamacodechi:
Yes but actually I've kept an eye on rollbar and this bill has not come up again and the timeout does not need to be raised so it's my opinion that we do not need to merge this at all and can close the issue. What do you think @clemenshelm ?

clemenshelm:
I agree 🙂

PR #375, opened 2017-09-01
321 vat number median height

### LINKED ISSUE #321: Vat Number Detection no longer requires the median height hack
It seems that with our ✨new✨clever✨Tesseract✨version ✨✨  the cause for the need to have median height taken into consideration (looking at you, Google invoices 👀 ) seems to be fixed!

Research with Google invoices and if this is indeed the case, remove the implementation and corresponding specs.

Tr4in:
Closes #321 

Works fine :+1:


clemenshelm:
Why did you remove the old spec?

Tr4in:
The old spec depends on median height which we don't need anymore due to the new Tesseract version. Now tesseract recognizes IE6388047V as a whole word on a google bill.

Tr4in:
The old tesseract recognized **IE** then **Rechnung** and then **6388047V** which isn't valid! Therefore we used the median height to remove the **Rechnung**. 

But now it recognizes IE6388047V as a whole word!

tamacodechi:
If I understand correctly then, you didn't really remove the spec but update it with data from the newest version of tesseract?

Tr4in:
right

tamacodechi (changes requested):
This looks very promising! 👍 
If I understand correctly, we do, in fact, not need the median height hack since the new version of tesseract picks up the vat ID as a single word. Why did you test it with a different file though? Can you just rewrite the data with the old file? this way we can know for sure this works for all files that required this weird quirky hack.

Tr4in:
On the old bill **wrj8fiNZQYjymoocT.pdf** the **number** after IE does not get recognized: 
selection_014

I can't update the old test because this **number** is missing(didn't get recognized)

tamacodechi:
Oh wow, that's so annoying! Have you checked with other google bills to see if this is a common occurrence? Let me know if you need more bill IDs!

Tr4in:
If you take a look at the list you will see that on the checked ones it gets recognized(I've tested every bill id)

PR #380, opened 2017-09-11
refactor detectors

### LINKED ISSUE #368: Evaluate and clean up regex accessing
Some of our detectors have adopted a strange way of dealing with regexes ie.
`/#{ALL_SYMBOLS.map { |s| Regexp.quote(s) }.join('|')}/`

This was the case in the relative date detector because in the calculation we do a check if any terms `.include?` *but* we can remove this and replace it with `=~` which instantly removes 3 lines of code and lots of complication (tested and proven to work!)

The detectors that use this style of filtering are:
- [ ] currency detector
- [x] invoice number label detector
- [x] relative date detector

I've tested removing it from the relative date detector and it works. The due date label detector contains the same logic but also works without it. Please investigate and save our detectors from unnecessary code! 👏

Tr4in:
Closes #368


tamacodechi:
👀 Currency?

lib/calculations/relative_date_calculation.rb
```diff
@@ -6,7 +6,6 @@ class RelativeDateCalculation
   def relative_date(date_relative_to)
     return nil if RelativeDateTerm.empty?
 
-    date_relative_to if
-      RelativeDateDetector::SAME_DAY_TERMS.include?(RelativeDateTerm.first.text)
+    date_relative_to if RelativeDateTerm.first.text =~ RelativeDateDetector::SAME_DAY_TERMS
```
tamacodechi, line 9: 🎉

tamacodechi (changes requested):
Wonderful, can you do the same for the currency detector? In the case where the text is matched try to stringify the term 😉

Issue #369, opened 2017-09-01
Find and filter out unlabelled invoice numbers at detection


tamacodechi:
Just as we did for the hungarian prices :tada: we should filter out unlabelled invoice numbers *at detection* so that our machine learning doesn't receive any false positive invoice numbers. We already save which invoice numbers should or should not be labelled, so there's a big hint :wink:

PR #381, opened 2017-09-11
moved logic to calculation

### LINKED ISSUE #370: False positive term deletion is a calculation, right?
So we should move all of this logic to the respective calculations, not detections. 

Calculations are always run in tandem with detections in our `.recognize` method so this should not have any weird side effects.

Tr4in:
Closes #370


lib/bill_recognizer.rb
```diff
@@ -242,6 +242,10 @@ def calculate_attributes(version)
   end
 
   def calculate_amounts
+    # First: Remove false positives
```
tamacodechi: No comment please 🙈 
If you want to make it super clear why you have these grouped, then make a separate method that runs these two like 
```ruby
def remove_false_positives
    PriceCalculation.remove_quantities
    PriceCalculation.remove_dates
end
```

The code should be so clear that it explains itself without comments 😉 

You could also have one method in the price calculation for removing false positives. Seeing as other terms may soon have this too #369 it may be good to have these defined in the calculations and then call the methods from one method before all the calculations. What do you think? 😊

tamacodechi (changes requested):
Great 🌮 Just some thoughts ✌️

tamacodechi:
Also please please the rubocop 🤖 💕

PR #383, opened 2017-09-13
Improve deployment

### LINKED ISSUE #261: Improve deployment
At the moment the deployment is a rake task. The problem with this is that all dependencies of all rake tasks are loaded which leads to some problems.

Please use a separate tool for deployment (like capistrano) to isolate it from the rake environment.

It would also be nice to commit, tag and push the newest version number created in the deployment. Make sure there are no uncommitted changes before deploying, though.

tamacodechi:
This branch doesn't follow our naming conventions 🙈 But that's because I have like 5 different branches on which I was testing and trying out how to improve deployment using different possible tools.

Closes #261 

I've chosen mina over Capistrano because it's so light and it does exactly what, to my understanding, we want. みんなありがとう :bowtie: 

> Jokes aren't funny once you explain them, but mina means everyone/all in Japanese. It says "Thank you everyone" AND "Thank you Mina" 🥁 ✨


tamacodechi:
Oh! Forgot to add! To deploy, simply run ```mina deploy``` 🌵

